PR #11, opened 2022-10-29
Add unit stats

cooperlees:
- Add new units module
- enable in default config ...
- If enabled in config, count units state stats from the dbus ListUnits call

Testing:
- Add unittest for parse unit (tuple returned from dbus)
- Run on a systemd linux box:
  - `cargo run -- -c monitord.conf -v`


src/json.rs
```diff
@@ -52,12 +53,50 @@ fn flatten_networkd(networkd_stats: &networkd::NetworkdState) -> HashMap<String,
...
+            "path_units" => Some(units_stats.path_units),
+            "scope_units" => Some(units_stats.scope_units),
+            "service_units" => Some(units_stats.service_units),
+            "slice_units" => Some(units_stats.slice_units),
+            "socket_units" => Some(units_stats.socket_units),
+            "target_units" => Some(units_stats.target_units),
+            "timer_units" => Some(units_stats.timer_units),
+            "total_units" => Some(units_stats.total_units),
+            _ => {
+                debug!("Got a unahndled stat '{}'", field_name);
```
flz, line 82: Typo.

PR #18, opened 2022-11-10
Replace networkctl exec with dbus call

### LINKED ISSUE #17: Remove networkctl exec for interface ID lookup
Find another way to convert interface id to interface name. Maybe (in order of preference):

- dbus
- netlink
- scrape ip command even ...

Up for other ideas readers ...

# dbus

Seems possible.

https://www.freedesktop.org/software/systemd/man/org.freedesktop.network1.html

cooperlees:
- Save the fork and exec
- 1 DBUS call to get all interface ID to name HashMap

Test:
- Fix tests
- Run on a Linux box: `cargo run -- -c monitord.conf`

Fixes #17


cooperlees:
Just got to fix unit tests tomorrow.

cooperlees:
Unit tests fixed! Merge time!

PR #55, opened 2023-09-21
Add base pid1 stats

### LINKED ISSUE #54: Monitord resource usage of PID 1 itself
Lets use procfs - Lets steal how simple-process-stats uses it:
- https://github.com/robotty/simple-process-stats/blob/master/src/linux.rs

cooperlees:
- Optionally (config driven) pull from procfs
  - https://manpages.debian.org/buster/manpages/procfs.5.en.html
- Goals:
  - See CPU, Memory + thread regressions & spikes
- Make new stats struct memory optional so we don't add to JSON if None

Test:
- Fix json tests + add a pid1 unittest
- Risky pid1 unitest expecting there always to be a PID1 - Should be :D

Fixes #54


README.md
```diff
@@ -164,6 +164,10 @@ Is semi pretty too + custom. All unittested ...
   "networkd.wg0.ipv6_address_state": 3,
   "networkd.wg0.oper_state": 9,
   "networkd.wg0.required_for_online": 1,
+  "pid1.cpu_time_kernel": 48,
+  "pid1.cpu_user_kernel": 41,
+  "pid1.memory_usage_bytes": 19165184,
+  "pid1.num_threads": 1,
```
DaanDeMeyer, line 170: From v254 onwards pid1 will never use more than a single thread so not sure this one is very useful. Maybe processes in the same cgroup could be useful to track?

DaanDeMeyer (approved):
LGTM but not sure about the threads

PR #80, opened 2024-07-17
Switch license file from LGPL 2.1 to GPL 2

michel-slm:
This matches `Cargo.toml`; the inconsistency was detected during the Fedora review process.


michel-slm:
License text can be checked against https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt linked to from https://spdx.org/licenses/GPL-2.0-only.html

cooperlees (approved):

cooperlees:
Thanks mate.

PR #83, opened 2024-09-05
Add EnumIter to SystemdActiveState and SystemdUnitLoadState

sergeyfd:
The change allows iterating over enums values which is really useful in some cases.


sergeyfd:
Updated. I don't think we need it in config.rs

cooperlees (approved):
Awesome - Thanks for adding to all!

PR #85, opened 2024-09-29
Move to zbus for all dbus communications

cooperlees:
- Move to async rust + zbus
- Generate new zbus traits into dbus module for respective systemd dbus buses
- Delete all dbus module usage + in Cargo.toml

This diff has not moved us to taking advantage of calling all dbus calls in parallel. Will do in dedicated diff if we decide to go forward here.

Pros:
- async
- More actively developed + liked zbus crate now used
- Drop libdbus dependency

Cons:
- 3.6MB vs. 1.4MB binary size
- More CPU used ... more than 3x "user" CPU time and 4x "sys" on same Fedora box reported using `time`


zeenix:
> 3.6MB vs. 1.4MB binary size

This is known issue and I've some concrete plans/ideas on fixing this (help welcome!): https://github.com/dbus2/zbus/issues/304#issuecomment-2144780012

> More CPU used ... more than 3x "user" CPU time and 4x "sys" on same Fedora box reported using time

Interesting. I'd not trust a simple `time` output comparison too much (you want to use something like `criterion` with `cargo bench`) but as I wrote on the Matrix channel, I recently removed the biggest performance bottleneck and the performance should now be a **lot** better. These changes aren't released yet but if you like to give it a try from our git main, that would be great.

PR #92, opened 2024-10-23
Replace auto discovery with allowlist and blocklist

lsiudut:
The change introduces allow and blocklist for container support, similar to the one used for units. Additionally a handly deserialization for `list_machiones` got introduced, using `serde_tuple`.


monitord.conf
```diff
@@ -34,7 +34,9 @@ sshd.service
 
 [machines]
 enabled = true
-auto_discover = true
```
cooperlees, line 37: So we've made the decision we're only doing auto discovery?

Cargo.toml
```diff
@@ -23,6 +23,7 @@ procfs = "0.15.1"
 serde = { version = "1.0.203", features = ["derive"] }
 serde_json = "1.0"
 serde_repr = "0.1"
+serde_tuple = "1.0.0"
```
cooperlees, line 26: This isn't in fbsource, but I hope that doesn't cause to much pain ...

I generally check fbsource before adding deps ... just cause of the fun that can bring ...

src/machines.rs
```diff
@@ -114,3 +128,40 @@ pub async fn update_machines_stats(
...
+            },
+            crate::dbus::zbus_machines::ListedMachine {
+                name: "baz".to_string(),
+                class: "container".to_string(),
+                service: "".to_string(),
+                path: OwnedObjectPath::try_from("/sample/object").unwrap(),
+            },
+        ];
+        let allowlist = vec!["foo".to_string(), "baz".to_string()];
+        let blocklist = vec!["bar".to_string()];
```
cooperlees, line 159: What's the behavior if someone puts bar in both lists? I believe it will still be blocked lookin at the filter function. Should we test that or throw an error if someone does put same name in both lists?

cooperlees (changes requested):
Look goods to me - Just wondering if we should 
- test a little more (see common on test for people defining in both lists)
- update readme that we only auto discover machiens based on the allow + deny list ....

monitord.conf
lsiudut: It doesn't really give us anything valuable. We can achieve exactly the same effect by not defining allowlist nor blocklist. If we define one of those though, we limit the resources to selected group. This is a matter of approach, this one mimics the logic from `units`.

Cargo.toml
lsiudut: Good point. I'll look into importing it.

src/machines.rs
lsiudut: I follow the logic from `units` file - https://github.com/cooperlees/monitord/blob/main/src/units.rs#L234 .

My intuitive approach was to make allowlist and blocklist are mutually exclusive - we either define a list of hosts to monitor or block certain entries and be happy with the remaining set. After inspecting the code though I decided to keep it consistent with existing allow/block list logic, simply to avoid confusion.

Cargo.toml
cooperlees: It should be fine. Third party for rust will actually bring it in for us I think here ... leave it to me.

src/machines.rs
cooperlees: If it's identical then we're good. I wonder if we should combine the code to select then so it will always be the same (another PR tho)

PR #96, opened 2024-11-07
Some very low-hanging binary-size & compile-time optimizations

zeenix:


zeenix:
> Thanks - Very nice.

No worries. If I may, I'd suggest not using squash-merging when the contributor has provided atomic commits with their own separate rationale and info. Up to you, of course. :)

PR #97, opened 2024-11-10
Some more tiny binary size reductions

zeenix:
More low-hanging fruit.


cooperlees (approved):
Awesome, thanks! I did try this but failed missing the dev-dependencies and 1 or 2 of the main features you've used.

Down to 2.3MB ... nice.

PR #113, opened 2025-08-20
Run collectors with a timeout

sergeyfd:
We ran into a situation when a dbus call `list_link` was silently hanging without producing any errors for `monitord`. Restarting `systemd_networkd` fixed that issue but it'd be nice to have some error message in the `monitord` logs indicating that dbus calls do to complete in time. While this diff doesn't do exactly that it adds timeout wrappers for all `collectors`. That will make them to either complete in time or produce a timeout error.


cooperlees (approved):
I dislike all the clones and code but I dislike sitting indefinately more ... so let's do this ... I'll do more changes once i merge. Just docs and stuff.

Issue #82, opened 2024-08-20
Collect dbus (itself) stats


cooperlees:
Add what makes sense as optional output for monitord
- Config enabled
  - When in daemon mode - Maybe also grab these stats less if the rerun period is < a dbus stat specific
    - The debug stats are non blocks / clostly to grab from dbus-broker for example

```
sudo dbus-send --system --dest=org.freedesktop.DBus --type=method_call --print-reply /org/freedesktop/DBus org.freedesktop.DBus.Debug.Stats.GetStats
```

ZBus has a module for this: https://docs.rs/zbus/latest/zbus/fdo/index.html
- Could enhance it's auto parsing to be friendlier
DBus Spec: https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-debug-stats-interface

### dbus-broker

- Would also love parsing these and making stats - especially quota usage
- Again, config enabled please

cooperlees:
Started work on this, but seems zbus doesn't like what dbus-broker returns:
- https://github.com/cooperlees/monitord/tree/dbus_stats

cooperlees:
https://github.com/dbus2/zbus/pull/1338 - Fix that will allow monitord to use fdo::get_stats from zbus

cooperlees:
The plot thickens, dbus-broker does not return the standard metrics: https://github.com/bus1/dbus-broker/issues/390
- So asking for that too :)

cooperlees:
@ikruglov has added these.

PR #116, opened 2025-10-02
implement collecting org.freedesktop.DBus.Debug.Stats.GetStats

ikruglov:
This PR implements collecting `org.freedesktop.DBus.Debug.Stats.GetStats` from dbus-broker.

Unfortunatelly, various DBus daemons (ex: dbus-broker and dbus-daemon) represent stats differently. Moreover, the stats vary across versions of the same daemon. 

The PR focused on dbus-broker stats and uses flexible approach providing max available information.

Need to discuss:
- peer names in metrics

TODO:
- tests
- allow/deny lists?

Sample output:
```
{
  "monitord.dbus.:1.0.outgoing_bytes": 0,
  "monitord.dbus.:1.1.outgoing_bytes": 0,
  "monitord.dbus.:1.2.outgoing_bytes": 0,
  "monitord.dbus.:1.3.outgoing_bytes": 0,
  "monitord.dbus.:1.5.outgoing_bytes": 0,
  "monitord.dbus.:1.75.outgoing_bytes": 0,
...
}
```

Implements https://github.com/cooperlees/monitord/issues/82

### DIFF
--- a/src/json.rs
+++ b/src/json.rs
@@ -676,6 +676,7 @@ mod tests {
                 .try_into()
                 .expect("Unable to make SystemdVersion struct"),
             machines: HashMap::from([(String::from("foo"), MachineStats::default())]),
+            dbus_stats: None,
         };
         let service_unit_name = String::from("unittest.service");
         stats.units.service_stats.insert(


src/dbus_stats.rs
```diff
@@ -0,0 +1,342 @@
...
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Default, Eq, PartialEq)]
+pub struct DBusBrokerPeerAccounting {
+    pub name: String,
+
+    // credentials
+    pub unix_user_id: Option<u32>,
+    pub process_id: Option<u32>,
+    pub unix_group_ids: Option<Vec<u32>>,
+    // ignoring LinuxSecurityLabel
+    // pub linux_security_label: Option<String>,
```
cooperlees, line 29: Let make note of this in the issue that someone else can implement this if needed

src/dbus_stats.rs
```diff
@@ -0,0 +1,342 @@
...
+        matches: get_u32(stats, "Matches"),
+        reply_objects: get_u32(stats, "ReplyObjects"),
+        incoming_bytes: get_u32(stats, "IncomingBytes"),
+        incoming_fds: get_u32(stats, "IncomingFds"),
+        outgoing_bytes: get_u32(stats, "OutgoingBytes"),
+        outgoing_fds: get_u32(stats, "OutgoingFds"),
+        activation_request_bytes: get_u32(stats, "ActivationRequestBytes"),
+        activation_request_fds: get_u32(stats, "ActivationRequestFds"),
+    });
+}
```
cooperlees, line 173:
```suggestion
   Some(DBusBrokerPeerAccounting {
        name: name.clone(),
        unix_user_id: get_u32(credentials, "UnixUserID"),
        process_id: get_u32(credentials, "ProcessID"),
        unix_group_ids: get_u32_vec(credentials, "UnixGroupIDs"),
        name_objects: get_u32(stats, "NameObjects"),
        match_bytes: get_u32(stats, "MatchBytes"),
        matches: get_u32(stats, "Matches"),
        reply_objects: get_u32(stats, "ReplyObjects"),
        incoming_bytes: get_u32(stats, "IncomingBytes"),
        incoming_fds: get_u32(stats, "IncomingFds"),
        outgoing_bytes: get_u32(stats, "OutgoingBytes"),
        outgoing_fds: get_u32(stats, "OutgoingFds"),
        activation_request_bytes: get_u32(stats, "ActivationRequestBytes"),
        activation_request_fds: get_u32(stats, "ActivationRequestFds"),
    })
}
```

Rust has implicit return in functions

src/dbus_stats.rs
```diff
@@ -0,0 +1,342 @@
...
+                "Bytes" => user.bytes = Some(pair),
+                "Fds" => user.fds = Some(pair),
+                "Matches" => user.matches = Some(pair),
+                "Objects" => user.objects = Some(pair),
+                _ => {} // ignore other fields
+            }
+        }
+    }
+
+    return Some(user);
```
cooperlees, line 281:
```suggestion
    Some(user)
```
Implicit return - Notice the lack of `;` here ...

src/dbus_stats.rs
```diff
@@ -0,0 +1,342 @@
...
+/// Async wrapper than can update dbus stats when passed a locked struct
+pub async fn update_dbus_stats(
+    connection: zbus::Connection,
+    locked_machine_stats: Arc<RwLock<MachineStats>>,
+) -> anyhow::Result<()> {
+    let mut machine_stats = locked_machine_stats.write().await;
+    match parse_dbus_stats(&connection).await {
+        Ok(dbus_stats) => machine_stats.dbus_stats = Some(dbus_stats),
+        Err(err) => error!("dbus stats failed: {:?}", err),
+    }
```
cooperlees, line 340: Lets get the machine_stats in the Ok(dbus_stats) clause
```suggestion
    match parse_dbus_stats(&connection).await {
        Ok(dbus_stats) => {
            let mut machine_stats = locked_machine_stats.write().await;
            machine_stats.dbus_stats = Some(dbus_stats)
        },
        Err(err) => error!("dbus stats failed: {:?}", err),
    }
```
This will ensure we hold the write lock for the shortest time possible and other concurrent collections can write asap

src/json.rs
```diff
@@ -401,6 +402,39 @@ fn flatten_machines(
...
+        let peer_name = &peer.name;
+        if let Some(ob) = peer.outgoing_bytes {
+            flat_stats.insert(
+                format!("{base_metric_name}.{peer_name}.outgoing_bytes"),
+                ob.into(),
+            );
+        }
+    }
+
+    flat_stats
```
cooperlees, line 435: Nice implicit return here :)

src/lib.rs
```diff
@@ -98,8 +101,8 @@ pub async fn stat_collector(
         // Always collect systemd version
 
         join_set.spawn(crate::system::update_version(
-            sdc.clone(),
-            locked_machine_stats.clone(),
+           sdc.clone(),
+           locked_machine_stats.clone(),
```
cooperlees, line 105: Wow, wonder how rustfmt missed this ...

cooperlees (changes requested):
Awesome - Thanks!

Let's get:
- Add actually read a config enable out of the ini config file
  - To do so I'll explain below
- `cargo fmt` formatting done
- Fix the `cargo clippy` errors
  - I commented inline on most

Feel free to add tests, but they can wait if you like for a dedicated PR and we can talk about what config options we want ... and do them in a dedicated diff now.

But I'm happy to do that once this is merged if you'd like in the interest of time and ge tthis released etc.

# Confi ini:
In the 
```
impl From<Ini> for Config {
    fn from(ini_config: Ini) -> Self {
```
on line 168, lets add something like:
```rust
        // [dbus] section
        config.dbus.enabled = read_config_bool(
            &ini_config,
            String::from("dbus"),
            String::from("enabled"),
        );
```
- You will also need ot add the `DbusConfig` to the main `Config` struct

Then I'm happy to merge. Thanks again!

src/dbus_stats.rs
ikruglov: do you want me to create an issue?

ikruglov:
@cooperlees, I've addressed your comment. Also, I've added the following:
- parsing dbus.enabled
- generating other metrics (see sample below)
- formating peer name ':1.0' -> '1-0'. Let me know if you're fine with this

IMO we should think about:
1. what metrics to produce (all or a subset). I think in practise we currently need only outgoing_bytes. But I'd produce more to have them if we need. @DaanDeMeyer, looking for your input.
2. filtering (relevant to 1). Do we do dbus_allow/denylist? What is it going to filter based on? There are peers and users.
3. dbus stats is daemon-specific unfortunately. Should we do something like `dbus.format = dbus-broker` in config file?

DaanDeMeyer:
I'll look in more detail tomorrow, but one quick comment is that we should map peers to service names (see busctl list) to get more descriptive metric names.

ikruglov:
one more update, I forgot to implement top-level metrics. Done now. These metrics seems to be generic across dbus daemons. 

```
  "monitord.dbus.active_connections": 6,
  "monitord.dbus.bus_names": 4,
  "monitord.dbus.incomplete_connections": 0,
  "monitord.dbus.match_rules": 15,
  "monitord.dbus.peak_bus_names": 4,
  "monitord.dbus.peak_bus_names_per_connection": 1,
  "monitord.dbus.peak_match_rules": 17,
  "monitord.dbus.peak_match_rules_per_connection": 8,
  ```

DaanDeMeyer:
> > we should map peers to service names (see busctl list) to get more descriptive metric names.
> 
> I've look into this already. It seems that peer names are not 1:1 mappable to human-firendly names. Example (multiple entries for :1.1, :1.2, etc):
> 
> ```
> [root@cebe55a14502 repo]# busctl list
> NAME                        PID PROCESS         USER            CONNECTION    UNIT                     SESSION DESCRIPTION
> :1.0                         67 systemd-resolve systemd-resolve :1.0          systemd-resolved.service -       -
> :1.1                         75 systemd-logind  root            :1.1          systemd-logind.service   -       -
> :1.2                          1 systemd         root            :1.2          init.scope               -       -
> :1.3                         77 dbus-broker-lau root            :1.3          dbus-broker.service      -       -
> :1.5                        737 systemd-network systemd-network :1.5          systemd-networkd.service -       -
> :1.86                     39659 busctl          root            :1.86         init.scope               -       -
> org.freedesktop.DBus          1 systemd         root            -             init.scope               -       -
> org.freedesktop.home1         - -               -               (activatable) -                        -       -
> org.freedesktop.hostname1     - -               -               (activatable) -                        -       -
> org.freedesktop.locale1       - -               -               (activatable) -                        -       -
> org.freedesktop.login1       75 systemd-logind  root            :1.1          systemd-logind.service   -       -
> org.freedesktop.network1    737 systemd-network systemd-network :1.5          systemd-networkd.service -       -
> org.freedesktop.oom1          - -               -               (activatable) -                        -       -
> org.freedesktop.portable1     - -               -               (activatable) -                        -       -
> org.freedesktop.resolve1     67 systemd-resolve systemd-resolve :1.0          systemd-resolved.service -       -
> org.freedesktop.systemd1      1 systemd         root            :1.2          init.scope               -       -
> org.freedesktop.timedate1     - -               -               (activatable) -                        -       -
> org.freedesktop.timesync1     - -               -               (activatable) -                        -       -
> ```

Indeed, but I would still use the human friendly name if available. Doesn't have to be the service name actually, could also be the name taken on the bus. Even org.freedesktop.network would be much better than the peer identifier.

DaanDeMeyer:
I think the registered name is also guaranteed to be the same across hosts which I don't think is guaranteed for the peer identifier.

cooperlees:
Once merged, I plan to "rustify" the code more. Ivan is newer to rust so working is fine here for now :)

DaanDeMeyer (approved):
Metrics output looks good, leaving code review to @cooperlees

ikruglov:
done. Sorry, I keep forgetting about them.

PR #117, opened 2025-10-08
Add dbus parsing tests

cooperlees:
Test the parsing works as we expect


cooperlees:
Going to try again ...

PR #127, opened 2025-12-03
dbus: copy cgroup name only if not present

noxiouz:
Instead of cloning cgroup_name to retrieve an Entry, use moved value and clone only if an Entry is vacant


noxiouz:
not sure if it's a noticeable win, but it seems reasonable to do

PR #131, opened 2025-12-04
perf: make flatten* return Vec

noxiouz:
family of flatten functions is used to extend BTreeMap. Extend accepts any Iter, not only BTreeMap. As extend() does not take advantage of sorted order and flatten keys are unique, constructing and returning Vec is more efficient.


cooperlees:
If it's definately improving, sure. I don't think there's a real reason I used btreemap to be honest.

Does the release build show any improvement in memory or runtime? We report how long it takes in logs.

PR #159, opened 2026-02-25
Followup fixes for #158

YapingLi04:
This addresses @ikruglov's comments on #158.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -177,8 +177,8 @@ pub async fn stat_collector(
                 if config_clone.varlink.enabled {
                     let socket_path = crate::varlink_units::METRICS_SOCKET_PATH.to_string();
                     match crate::varlink_units::update_unit_stats(
-                        config_clone,
-                        stats_clone,
+                        Arc::clone(&config_clone),
+                        stats_clone.clone(),
                         socket_path,
                     )
                     .await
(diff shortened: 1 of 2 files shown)


src/varlink/metrics.rs
```diff
@@ -62,18 +62,14 @@ impl ListOutput {
         self.object.as_deref().unwrap_or("").to_string()
     }
 
-    /// Returns the string value or default_value if not present
-    pub fn value_as_string<'a>(&'a self, default_value: &'a str) -> &'a str {
-        self.value.as_str().unwrap_or(default_value)
+    /// Returns the string value if present
+    pub fn value_as_string(&self) -> Option<&str> {
+        self.value.as_str()
     }
```
ikruglov: still, value is a mandatory field in a metric. I think you should be able to remove Option<>. Maybe you can validate metric structure before parsing it

src/lib.rs
```diff
@@ -171,12 +171,29 @@ pub async fn stat_collector(
...
+                    .await
+                    {
+                        Ok(()) => Ok(()),
+                        Err(err) => {
+                            warn!(
+                                "Varlink units stats failed, falling back to D-Bus: {:?}",
+                                err
+                            );
+                            crate::units::update_unit_stats(config_clone, sdc_clone, stats_clone)
+                                .await
```
ikruglov: you calling `crate::units::update_unit_stats` here and a couple of lines below. Maybe merge them.

pseudo code:

```
if (varlink_enabled) {
  call_varlink()
  if success
     return

  log("Fallback to DBus")
}

call_dbus()
```

PR #171, opened 2026-03-11
[monitord] Only discard D-Bus connections on transport-level errors

ikruglov:
Not all errors indicate a broken connection. Request-level errors (MethodError, FDO, etc.) leave the connection alive. Only evict cached connections and signal unusable host connection on I/O or handshake failures.


cooperlees:
I don't think we'll need this.

PR #172, opened 2026-03-16
Add new metrics added in systemd

YapingLi04:
Add support for the metrics added in https://github.com/systemd/systemd/pull/41078

### DIFF
--- a/src/json.rs
+++ b/src/json.rs
@@ -71,6 +71,14 @@ fn flatten_networkd(
             format!("{interface_base}.required_for_online"),
             (interface.required_for_online as u64).into(),
         ));
+        flat_stats.push((
+            format!("{interface_base}.required_oper_state_for_online_min"),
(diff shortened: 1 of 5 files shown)


cooperlees (commented):
I need to look at this in depth. In future can we split (even in commits) units, networkd and each modules changes if possible to allow for easier review?

cooperlees (changes requested):
Should this be ready for cleaning up and looking to merge now? Let's also ask AI to add all the new metrics into README.md's example flat JSON output

Are these varlink only metrics? If so, we might need to start marking some metrics as varlink only in docs ...

cooperlees:
Probably easier to do this from scratch again. I'll probably do one tonight as I want to turn on varlink at Meta. And we're going to be close to varlink for all. Closer than I expected so nice work all.

PR #182, opened 2026-05-13
Support MONITORD_CONFIG env var for config path

aleivag:
Allow setting the config file path via the MONITORD_CONFIG environment variable as a fallback between the -c CLI flag and the default /etc/monitord.conf. Uses clap's built-in env feature for proper precedence handling and --help display.


cooperlees:
Yeah I thought rustfmt would be unhappy here ...

Issue #183, opened 2026-05-13
Varlink path emits thousands of zero-valued service_stats keys not present in D-Bus path


aleivag:
When using the varlink collection path (`[varlink] enabled = true`), the varlink endpoint exposes a few extra metrics that the dbus interface does not, `NRestarts` being one them. 

The `NRestarts` metric handler in `varlink_units.rs` creates a full `ServiceStats` entry (16 fields) for every service unit via `.or_default()`, even though only `nrestarts` is populated. The remaining 15 fields (CPU, memory, IO, timestamps, etc.) are all zeros.

On a system with ~390 services, this produces ~6,270 extra JSON keys — nearly all zeros — that the D-Bus path does not emit. The D-Bus path only creates `ServiceStats` entries for services explicitly listed in the `[services]` config section, and populates all fields with real data from D-Bus property calls.

### Impact

- **3x output bloat**: total JSON keys jump from ~2,800 (D-Bus) to ~9,000 (varlink), with ~6,270 being useless zeros
- **Misleading data**: consumers see `service_stats` entries and may assume the values are real, but only `nrestarts` is populated
- **Inconsistent behavior**: the same config produces different output depending on whether varlink is enabled

### Root cause

`varlink_units.rs` line 140-144:
```rust
"NRestarts" => {
    ...
    stats.service_stats.entry(object_name.to_string()).or_default().nrestarts = nrestarts;
}
```

The `or_default()` creates a full `ServiceStats` with all fields zeroed. The varlink path has no access to the `[services]` config and does not gate this on the services allowlist.

### what i would suggest

now i dont mind the nrestart metric because its true , its the other 0 that bugs me because they are not true, so what i would suggest is make SystemdUnitStats fields Option<u64>  and only serialize to json when this field is true

cooperlees:
Hmm, I wonder how / why the integration test doesn't see this. Does json-flat output normailze this somehow? Please try and enhance it in the PR who ever takes fixing this on.

cooperlees:
Fixed in #199 (ef4ce19d338477b2d97e5750923aec01ef20006b, 2026-09-17).

## What changed

Your diagnosis was exactly right — the handler gated on the wrong thing. It used the `state_stats` allowlist, which is empty by default and therefore matches every unit, so `.or_default()` minted a full 16-field `ServiceStats` for every service on the box:

```rust
 "NRestarts" => {
-    if !config.state_stats || should_skip_unit(&object_name, config) {
[...]

PR #184, opened 2026-05-13
Make ServiceStats fields nullable to avoid zero-filled output (#183)

aleivag:
The varlink path creates ServiceStats entries for every service unit via NRestarts metrics, but only populates nrestarts — the remaining 15 fields were emitted as zeros. On a system with ~390 services this produced ~6,270 useless  and false JSON keys.

Making all ServiceStats fields Option<T> with skip_serializing_if means only fields with actual data are emitted. The D-Bus path wraps all values in Some(), so its output is unchanged. The varlink path now emits only nrestarts per service instead of 16 fields.

now we only produce 3224 metrics but they all are true

### DIFF
--- a/src/json.rs
+++ b/src/json.rs
@@ -647,22 +647,7 @@ mod tests {
   "pid1.fd_count": 69,
   "pid1.memory_usage_bytes": 69,
   "pid1.tasks": 1,
-  "services.unittest.service.active_enter_timestamp": 0,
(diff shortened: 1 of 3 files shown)


cooperlees (changes requested):
As discussed, I think I'd rather fix or revert the change for this to try and use varlink if we can't fix it in the near term and just use dbus ...

PR #185, opened 2026-05-15
Add unit_files stats for generated and transient unit file counts

aleivag:
## Summary
- Add `UnitFilesStats` and `UnitFilesScope` structs to track unit file counts by scope (root/user) and unit type
- Count generated unit files from `/run/systemd/generator{,.early,.late}` and transient unit files from `/run/systemd/transient` and `/run/user/*/systemd/transient/`
- Filesystem-based counting — no D-Bus `ListUnitFiles` or varlink overhead
- Container-aware via `fs_root` parameter (`/proc/<pid>/root/...` prefix)
- Consistent across D-Bus and varlink collection paths

### Metric keys
```
monitord.unit_files.root.generated.<type>_units
monitord.unit_files.root.transient.<type>_units
monitord.unit_files.user.transient.<type>_units
```

Where `<type>` is the unit file extension (service, mount, swap, socket, scope, etc.)

## Test plan
- [x] `cargo test --lib` — 107 tests pass
- [x] New tests for `count_unit_files_by_type` (files only, not dirs), nonexistent dir (returns empty), and end-to-end `collect_unit_files_stats` with temp dir
- [x] Verified output with `cargo run -- -c monitord.conf` shows correct counts

### DIFF
--- a/src/units.rs
+++ b/src/units.rs
@@ -55,6 +55,22 @@ pub struct UnitsCollectionTimings {
     pub service_dbus_fetches: u64,
 }
 
+/// Unit file counts for a scope (root or user), broken down by unit type.
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Default, PartialEq)]
+pub struct UnitFilesScope {
+    /// Generated unit files by type (e.g. "service" => 2, "mount" => 5)
+    pub generated: HashMap<String, u64>,
+    /// Transient unit files by type (e.g. "service" => 10, "scope" => 6)
+    pub transient: HashMap<String, u64>,
+}
+
+/// Unit file statistics collected from the filesystem for root and user scopes.
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Default, PartialEq)]
+pub struct UnitFilesStats {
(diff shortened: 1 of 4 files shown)


src/units.rs
```diff
@@ -429,6 +440,30 @@ fn parse_unit(stats: &mut SystemdUnitStats, unit: &ListedUnit) {
...
+}
+
+/// Collect unit file statistics from the filesystem.
+pub fn collect_unit_files_stats() -> UnitFilesStats {
+    UnitFilesStats {
+        generated_unit_files: count_files_in_dir(GENERATOR_DIR),
+        transient_unit_files: count_files_in_dir(TRANSIENT_DIR),
+    }
+}
+
```
cooperlees, line 518: I'll merge but see if we can do in parallel via tokio threads too

cooperlees (approved):
LGTM - way cheaper than calling systemd via RPC.

cooperlees (changes requested):
copilot is right. Let's make varlink work + async parallel the calls + make it work for machined containers or if that's too much work just state it doesn't work for machines and open an issue.

cooperlees (approved):

PR #191, opened 2026-06-08
Update strum from 0.27 to 0.28

dtolnay:


cooperlees:
Is it better I turn dependabot back on and always be up to date?

dtolnay:
Up to you, sending my own PRs doesn't make a difference to me.

Keeping `strum` updated is relevant (like in https://github.com/cooperlees/monitord/pull/109) because it is used by fbmonitord.

```console
Action failed: fbcode//metalos/facebook/fbmonitord:fb_monitord_lib (rustc check)
error[E0599]: no variant or associated item named `iter` found for enum `SystemdUnitActiveState` in the current scope
   --> fbcode/metalos/facebook/fbmonitord/src/fb303_stats.rs:643:42
    |
643 |     for state in SystemdUnitActiveState::iter() {
    |                                          ^^^^ variant or associated item not found in `SystemdUnitActiveState`
    |
note: there are multiple different versions of crate `strum` in the dependency graph
   --> third-party/rust/vendor/strum-0.27.2/src/lib.rs:100:1
    |
100 | pub trait IntoEnumIterator: Sized {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this is the expected trait
    |
   ::: third-party/rust/vendor/strum-0.28.0/src/lib.rs:99:1
    |
 99 | pub trait IntoEnumIterator: Sized {
    | --------------------------------- this is the trait that was imported
    = help: you can use `cargo tree` to explore your dependency tree
```

cooperlees:
Yeah I get it, I just turned of dependabot to stop the pain the other way ... wondering if it's better to live with that now and then we can just cut new versions when we need to pull internally?

dtolnay:
Makes sense, I am fine with that.

PR #194, opened 2026-06-24
Track stale dbus-broker fds

aleivag:
This will keep track of stale_pids leave in dbus... this seems to be an issue when some process dont close their connection gracefully ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,8 @@ output_format = json
 [dbus]
 # Summary counters - both dbus-broker + dbus-daemon
 enabled = false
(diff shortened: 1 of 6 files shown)


cooperlees:
Docs + config updates too please. Happy to have this on by default, but let's allow people to turn it off incase it's heavy/buggy etc.

I also need to see if this is as parallel / async as it could be ... Please ask you clanker to maintain that cause it seems serial here where the dbus and file system operations could be done in parallel from a quick look ...

PR #198, opened 2026-09-12
Keep D-Bus proxies async-only when zbus blocking support is enabled

rrnewton:
When another dependency enables `zbus/blocking-api`, monitord's proxy macros also generate blocking methods. The Manager proxy's object-returning methods then fail to compile because `MachineProxyBlocking` is not in scope.

Keep all seven crate-private D-Bus proxies explicitly asynchronous, with a reminder to preserve that setting when regenerating them. Add a CI compile check with the dependency's blocking feature enabled to cover feature unification.

Validation with Rust 1.96.0 and zbus 5.19.0:

- `cargo check --all-targets --features zbus/blocking-api`: fails on unchanged upstream `2b00638d` with E0425/E0433; passes with this change.
- `cargo test`: 114 passed, zero failed or ignored.
- `cargo fmt --check`: passed.
- `cargo clippy`: passed.
- `cargo build --release --all-features`: passed.


cooperlees (approved):
Sure thing, this is what I wanted long ago. Is this a newer feature option?

Anyways, thanks!

Issue #37, opened 2023-07-13
Move to varlink if/when APIs are available


cooperlees:
There is world that the dbus calls we use are possibly moving to be exposed via varlink on PID 1 itself. If that happens lets use it as there are less dependencies and cheaper lookups etc. etc.

- The first few versions of monitord will support both and config will drive which is used
- Eventually we'll remove dbus support once a version of systemd with varlink is mainstream

Experiment using async varlink client:
- https://github.com/zeenix/zlink

cooperlees:
Looking today at systemd APIs, still seems we need a lot more implemented before monitord can go anywhere near varlink
- There is also still no async/tokio client

```
[root@b89a51f4c266 systemd]# find . | grep 'io\.'
./userdb/io.systemd.DynamicUser
./userdb/io.systemd.Multiplexer
./userdb/io.systemd.NameServiceSwitch
./userdb/io.systemd.DropIn
./userdb/io.systemd.Home
./io.systemd.ManagedOOM
./io.systemd.Credentials
./journal/io.systemd.journal
./oom/io.systemd.ManagedOOM
./io.systemd.BootControl
./io.systemd.sysext
./netif/io.systemd.Network
./resolve/io.systemd.Resolve
./resolve/io.systemd.Resolve.Monitor
./io.systemd.Hostname
```
- This is with a pretty new systemd too (257.3-7.fc43)

Going to park it for awhile.

cooperlees:
https://github.com/systemd/systemd/issues/36755 - Posted asking for more documentation and a way to see when the APIs we need are available etc. etc.

cooperlees:
https://github.com/systemd/systemd/issues/36877 - Asking for new varlink API to move off parsing the private networkd files we parse ...

cooperlees:
## Remaining varlink work (follow-up to #199–#202)

### Left in monitord — small (same pattern as #200)

_Validated against Fedora Rawhide (systemd 262~rc3) with the varlink-vs-D-Bus parity test, rewritten in Python and runnable locally in #203._

- [x] `UnitsByLoadStateTotal` (v261+) for loaded/masked/not-found instead of per-unit parsing — #204
- [x] `StatusErrno` per-service metric (v261+, feeds full `ServiceStats`) — #205
[...]

PR #203, opened 2026-09-18
Rewrite the varlink integration test in Python so it runs locally

cooperlees:
Prep work for the remaining [#37](https://github.com/cooperlees/monitord/issues/37) varlink items: each of those changes needs a varlink-vs-D-Bus parity run to validate it, and that check was ~140 lines of inline bash in the Action — reproducing a CI failure meant hand-replaying docker commands.

## Changes

- `varlink_integration_test.py` — manages the Fedora Rawhide container itself (build image, boot systemd, build monitord, run both paths, compare). Reuses a running container between invocations for fast iteration; `--fresh` recreates it.
- `varlink_integration_unittests.py` — unit tests for the pure logic (config generation, output comparison, fallback detection).
- `.github/workflows/varlink-integration.yml` — now just runs the two scripts.

## Behaviour changes vs the bash version

- **New assertion:** the varlink run must not log a D-Bus fallback. Previously, if the varlink socket was missing or a collector errored, every collector fell back to D-Bus and the comparison passed vacuously by diffing D-Bus output against itself.
- Comparison reports per-key differences (`key: dbus=X varlink=Y`) instead of a `diff -u` of filtered lines.
- Builds into a container-local `CARGO_TARGET_DIR=/ctarget`, so running this on a developer machine does not overwrite `target/` with Linux artifacts.
- Same fixture units, same excluded keys, same time-in-state assertion as before.

## Verification

Run locally against Fedora Rawhide (systemd 262~rc3-1.fc46):

```
=== Asserting the varlink run used varlink ===
PASS: no collector fell back to D-Bus
=== Asserting time-in-state is populated on both paths ===
PASS: varlink time_in_state_usecs=332940346
PASS: dbus time_in_state_usecs=332818630
=== Comparing D-Bus and varlink output ===
PASS: D-Bus and varlink outputs are identical (59 keys compared)
```

`python3 varlink_integration_unittests.py`: 12 passed.


cooperlees:
Addressed all five Copilot comments in dec0718.

**Both vacuous-pass holes were real** — these are the important ones, since either would let the parity check pass while comparing nothing:

- **varlink toggle not verified** — `build_ci_configs` now fails if there was no `[varlink] enabled = false` to flip. Otherwise both configs select the same path and the comparison diffs a D-Bus run against itself. (`assert_no_varlink_fallback` would not catch it: with varlink disabled, nothing logs a fallback.)
- **allowlist substitution checked globally** — now tracked per substitution as it happens. The fixture units also appear in `[services]`, so the old whole-config search would pass with an empty `state_stats` allowlist, and the run would compare no per-unit state.
- **unbounded section slice in the unit test** — `section_body()` now bounds at the next `[` header.

**Container reuse by name** — fixed. `container_matches()` checks the running container's image and its `/workspace` bind source before reuse, and replaces it otherwise. Worth noting Docker Desktop reports bind sources under a `/host_mnt` prefix, so the comparison strips that; without it the check would fail on macOS and recreate the container on every run.

**`docker rm -f` on a missing container** — the premise doesn't reproduce on current Docker:

```
$ docker rm -f definitely-not-a-real-container-xyz
exit=0
```

Docker 29.8.0 treats it as a no-op. But it does error on older versions and on podman, so it is now guarded anyway — cheap, and it removes the version dependence.

Three new unit tests cover the new failure modes (14 total, all passing). Re-ran the full integration test against Fedora Rawhide (systemd 262~rc3): container correctly reused, no fallback to D-Bus, outputs identical across 59 compared keys.

PR #205, opened 2026-09-18
Consume the StatusErrno varlink metric

cooperlees:
Second of the small remaining varlink items from the [#37](https://github.com/cooperlees/monitord/issues/37) audit, same pattern as #200.

## Change

`ServiceStats::status_errno` was always 0 on the varlink path — the D-Bus path reads it from `org.freedesktop.systemd1.Service`, but nothing parsed it from the metrics. systemd v261+ exposes `io.systemd.Manager.StatusErrno` per service, so consume it alongside `NRestarts`, under the same `[services]` gating.

Negative values are preserved rather than rejected: systemd reports service failures as a negative errno, and the D-Bus property behaves the same way.

## Verification

Live on Fedora Rawhide (systemd 262~rc3-1.fc46), using a service that reports a real errno via `systemd-notify ERRNO=13`:

```
systemctl show errno-test.service -p StatusErrno --value   ->  13

before (varlink): {'errno-test.service.status_errno': 0}
after  (varlink): {'errno-test.service.status_errno': 13}
after  (D-Bus):   {'errno-test.service.status_errno': 13}
```

- `cargo test`: 123 passed (new `test_parse_one_metric_status_errno` covers the negative value and the `[services]` gating).
- `cargo fmt --check` and `cargo clippy`: clean.
- Varlink integration test on Rawhide: D-Bus and varlink outputs identical, no collector fell back to D-Bus.

Note: touches the same README requirements bullet as #204, so whichever lands second may need a trivial rebase.


cooperlees:
Good catch — the comment was backwards. Fixed in a43213c, and rebased onto main now that #204 has landed.

Confirmed on systemd 262 rather than taking it from the type signature. systemd refuses a negative `ERRNO=` outright:

```
$ systemd-notify ERRNO=-13
systemd[1]: errno-neg.service: Failed to parse ERRNO= field value '-13' in
            notification message: Numerical result out of range
$ systemctl show errno-neg.service -p StatusErrno --value
0
```

while `ERRNO=13` reports `13` on both the D-Bus property and the varlink metric. So the value is non-negative, as you said.

Changes:
- The comment now states the value is non-negative and explains why it is still stored as `i32` — that is the type of the D-Bus `StatusErrno` property `ServiceStats` is modelled on, which is what makes the conversion worth range-checking at all.
- The test fixture uses `13` (a value systemd can actually produce) instead of `-69`.
- Added a case for an out-of-range value, since an unsigned wire value is the realistic way to overflow `i32`: it warns and leaves the previous value rather than wrapping.

`cargo test`: 124 passed. Integration test re-run on Rawhide after the rebase: outputs identical, no fallback to D-Bus.

PR #206, opened 2026-09-18
Collect systemd version and system state over varlink

cooperlees:
Third of the small remaining varlink items from the [#37](https://github.com/cooperlees/monitord/issues/37) audit: *"`SystemState` + `Version` metrics (or `Manager.Describe`) to kill two D-Bus collectors"*.

## Change

The `version` and `system_state` collectors now try `io.systemd.Manager.Describe` on PID 1's varlink socket (v258+) and fall back to the D-Bus properties, the same shape the networkd and units collectors already use. Containers do the same through `/proc/<leader_pid>/root/run/systemd/io.systemd.Manager`.

New: `src/varlink/manager.rs` (proxy, declaring only the two runtime fields we consume so the rest of the ~3KB reply deserializes away) and `src/varlink_system.rs`.

`system.rs`'s state-string match moved into `parse_system_state()` so both paths share one mapping.

## Why Describe and not the metrics

The `Version`/`SystemState` metrics on `/run/systemd/report/io.systemd.Manager` would also work, but `io.systemd.Metrics` only offers an unfiltered `List` — no name filter (confirmed by introspection) — so reading two values means streaming every metric for every unit: **~11ms** measured, against Describe's single ~3KB reply.

## Verification

Live on Fedora Rawhide (systemd 262~rc3-1.fc46) — both paths agree exactly:

```
D-Bus:   system-state=3 version=262.rc3-1.fc46
varlink: system-state=3 version=262.rc3-1.fc46
```

Collector cost (3 runs each, ms):

| collector | D-Bus | varlink |
|---|---|---|
| `version` | 0.33–0.41 | 0.41–0.55 |
| `system_state` | 0.28–0.38 | 3.94–4.57 |

The two Describe calls are served serially by PID 1, so the second one waits on the first — that is the ~4ms. End-to-end run time is unchanged: the units collector still gates the cycle at ~15ms. A single shared Describe feeding both collectors would remove it, at the cost of coupling the two; happy to do that as a follow-up if you'd prefer it.

- `cargo test`: 126 passed (new tests: Describe reply deserialization incl. unknown fields, version string parity with the D-Bus property, and the state mapping).
- `cargo fmt --check` and `cargo clippy`: clean.
- Varlink integration test on Rawhide: outputs identical, and the "no collector fell back to D-Bus" assertion now also covers these two.


cooperlees:
Addressed in 0f86e4c, rebased onto main now that #203/#204 have landed.

**"A broken host Describe call could fall back to D-Bus while CI still passes"** — this one doesn't hold, and I tested it rather than arguing from the code. The integration test's `assert_no_varlink_fallback` greps the varlink run's log for fallbacks, and both new collectors log one. Pointing `MANAGER_SOCKET_PATH` at a nonexistent socket and re-running:

```
=== Asserting the varlink run used varlink ===
W0918 src/lib.rs:325] Varlink system state failed, falling back to D-Bus: I/O error: No such file or directory (os error 2)
W0918 src/lib.rs:255] Varlink version failed, falling back to D-Bus: I/O error: No such file or directory (os error 2)
FAIL: varlink run fell back to D-Bus (see above)

script exit code = 1
```

So CI fails, and names which collector regressed. (That assertion is new in #203 — it was added precisely because comparing final output after fallback proves nothing.)

**The socket check did miss the new socket** — fixed, it now checks `/run/systemd/io.systemd.Manager` alongside the metrics and networkd sockets.

**The container path is genuinely untested** — you're right, and it is not specific to this PR: CI runs no systemd-nspawn machine, so the entire `machines` collector is unexercised, including the container varlink paths added in #201 and #202 (`collector_timings.machines.success = 0` in the container run). Adding an nspawn fixture is worth doing, but it is a separate piece of work rather than something to bolt onto this PR — happy to file it against #37 if you want it tracked.

**Suppressed comment re: "Installed systemd version"** — correct, fixed. `Describe` reports the running manager's version, which can trail the installed package until PID 1 re-execs, and the README now says so.

`cargo test`: 127 passed. Integration test on Rawhide: all three sockets present, no fallback, outputs identical.

PR #208, opened 2026-09-18
Stop the varlink version and system state collectors queueing on PID 1

### LINKED ISSUE #207: varlink Describe calls queue ~10ms behind the units List stream on PID 1
Follow-up from #206, where I noticed the varlink `system_state` collector was ~10x slower than the D-Bus one and said in the PR that PID 1 was serialising the two `Manager.Describe` calls. **That explanation was wrong** — I measured it properly afterwards and the cost is ours, not systemd's. Filing what I found so it can be fixed separately.

## What it looks like

One-shot runs on Fedora Rawhide (systemd 262~rc3), 5 runs each, collector `elapsed_ms`:

| collector | D-Bus | varlink |
|---|---|---|
| `version` | 0.17–0.36 | 0.40–4.07 |
| `system_state` | 0.17–0.36 | 3.56–4.01 |

Daemon mode (`daemon_stats_refresh_secs = 1`), all collectors enabled, varlink on:

```
cycle 1: version=0.48  system_state=0.58   units=10.25  networkd=1.08  total=10.32
cycle 2: version=9.99  system_state=9.45   units=16.02  networkd=1.59  total=16.18
cycle 3: version=10.87 system_state=10.55  units=16.31  networkd=3.32  total=16.63
cycle 4: version=1.16  system_state=9.93   units=15.87  networkd=2.02  total=16.05
cycle 5: version=9.67  system_state=9.97   units=15.93  networkd=2.06  total=16.25
cycle 6: version=9.89  system_state=10.33  units=16.05  networkd=2.45  total=16.50
```

So it is not confined to `system_state`: in steady state every varlink collector inflates, `units` included (10.25 → ~16ms).

## What it is not

- **Not PID 1 being slow.** A raw `varlinkctl call /run/systemd/io.systemd.Manager io.systemd.Manager.Describe` is ~1ms wall *including process startup*.
- **Not the two Describe calls contending.** With `[system-state] enabled = false`, so only one Describe in the cycle, `version` is still bimodal: `0.52 0.50 4.33 0.56 3.81`.
- **Not leaked connections.** PID 1's fd count is flat at 123 before, during and after a 6-cycle daemon run.

## What it is

Concurrency between varlink collectors. With every other collector disabled, the `version` collector alone stays flat across daemon cycles:

```
cycle 1: 1.04   cycle 2: 0.77   cycle 3: 2.07   cycle 4: 1.89
cycle 5: 1.45   cycle 6: 1.66   cycle 7: 1.92
```

Same call, same socket — ~1-2ms alone vs ~10ms when the other collectors run alongside it.

The likely cause is that every varlink call builds its own executor. `varlink_units::collect_metrics`, `varlink_networkd::get_networkd_state` and `varlink_system::describe_runtime` each do:

```rust
tokio::task::spawn_blocking(move || {
    let rt = tokio::runtime::Builder::new_current_thread().enable_all().build()?;
    rt.block_on(async move { ... })
})
```

which is there because the zlink connection is `!Send` and cannot be held across an `await` in a `Send` future. So a cycle constructs and tears down one current-thread runtime per varlink call — four of them today (units, networkd, version, system_state), concurrently, and one more per call per container.

I have not confirmed the mechanism beyond this, and one observation does not fit neatly: in the daemon run above, cycle 1 is the *fast* one, which is the opposite of thread-pool warmup.

## Possible directions

1. **One shared `Describe` for `version` + `system_state`.** Halves the calls; this was the original suggestion in #206. Helps, but does not address the root, and couples the two collectors (they have separate config gates and separate `collector_timings` entries).
2. **One long-lived varlink executor** — a dedicated thread with a `LocalSet`, or a single reused current-thread runtime that all varlink calls dispatch onto — instead of building one per call. Addresses the root and helps `units` and `networkd` too.
3. Check whether zlink can be driven directly from the main runtime without `spawn_blocking`, which would remove the pattern entirely.

## Impact

No effect on end-to-end time today: the `units` collector gates the cycle either way, and the whole run is ~16ms. This is worth fixing for what it implies rather than what it currently costs — the per-call overhead scales with the number of varlink collectors and with the number of containers, and #37 is moving more collectors onto varlink.

Refs #37, #206

cooperlees:
Fixes #207 — the regression #206 introduced, where `version` and `system_state` went from ~0.3ms on D-Bus to ~10ms on varlink.

## Two causes

**1. PID 1 serves varlink requests one at a time.** The second `Manager.Describe` sat behind the units collector's entire `io.systemd.Metrics.List` stream (~15ms of PID 1 time). Proof: with `[units] enabled = false`, both Describes were ~1-2ms; with units on, one or both were ~10ms.

Both collectors now await a **single shared `Describe` per cycle** (`futures_util::future::Shared`). The first to poll drives the call, the other waits on the same result — so PID 1 is occupied once, not twice, and the collectors keep their separate config gates, fallbacks and `collector_timings` entries. Containers get one shared call each.

**2. `spawn_blocking` was not needed for these calls.** The `!Send` justification holds only for the streaming `List` call. Verified by compiling `assert_send` against each future:

| call | direct future `Send`? |
|---|---|
| `io.systemd.Manager.Describe` | **yes** |
| `io.systemd.Network.Describe` | **yes** |
| `io.systemd.Metrics.List` | no — `#[zlink(more)]` returns `dyn Stream` with no `+ Send` |

So `varlink_system` and `varlink_networkd` no longer build a current-thread tokio runtime per call. `varlink_units` is untouched and keeps the wrapper, with the comment corrected to say why it specifically needs it.

## Measurements (Fedora Rawhide, systemd 262~rc3)

Daemon mode, 1s interval, all collectors, varlink enabled:

```
before  cycle 2: version=9.99  system_state=9.45   units=16.02  total=16.18
        cycle 3: version=10.87 system_state=10.55  units=16.31  total=16.63
        cycle 5: version=9.67  system_state=9.97   units=15.93  total=16.25

after   cycle 2: version=1.75  system_state=1.63   units=15.69  total=16.02
        cycle 3: version=1.53  system_state=1.43   units=15.50  total=15.83
        cycle 5: version=1.50  system_state=1.37   units=16.17  total=16.52
```

One-shot, 5 runs, now level with D-Bus:

```
D-Bus:   v=0.43 ss=0.40  v=0.28 ss=0.28  v=0.27 ss=0.25  v=0.25 ss=0.35  v=0.35 ss=0.34
varlink: v=0.34 ss=0.31  v=0.45 ss=0.41  v=0.40 ss=0.37  v=0.33 ss=0.31  v=0.33 ss=0.30
```

End-to-end run time is unchanged — `units` gates the cycle either way — but each varlink collector added under #37 would otherwise join the same queue.

## Verification

- `cargo test`: 129 passed. New test asserts a failed shared `Describe` reaches both awaiters, so neither can report success off a call that never happened.
- Varlink integration test on Rawhide: no collector fell back to D-Bus, outputs identical across 59 keys.
- Confirmed the shared future is built per cycle, not once per process: version and system state stay correct across daemon cycles.


cooperlees:
The ordering observation is correct, and I could reproduce it — but I am not taking the suggested fix. Detail, since this was worth measuring rather than arguing:

**The race is real.** Under 8 CPU hogs on a 10-core box, the shared `Describe` does land behind the metrics stream on some cycles:

```
cycle 1: version= 0.62  system_state= 0.00   cycle 6: version= 5.98  system_state= 5.88
cycle 4: version= 9.01  system_state= 8.95   cycle 9: version=11.45  system_state=10.92
```

(`system_state=0.00` on cycle 1 is the sharing doing its job — the result was already resolved when the second collector awaited it.)

**I tried fixing the ordering without gating.** The non-coupling version of your suggestion is to drive the shared future from its own task, so the request goes out before the units collector is even spawned. Implemented and A/B'd it under identical load:

| | slow cycles (>5ms) | median |
|---|---|---|
| with eager drive | 4/10 | 4.80ms |
| lazy (as submitted) | 5/10 | 5.84ms |

Indistinguishable. Under CPU starvation the ordering is decided by the scheduler, not by when the future is first polled, so the extra machinery bought nothing and I dropped it.

**I am not gating the units collector on this call.** That would make the collector which determines cycle wall time wait on an unrelated socket — and a stalled or slow manager socket would then delay the whole cycle instead of just the two collectors that need it. That trade goes the wrong way for a call whose worst case is already bounded.

**The worst case here is the old behaviour, for one call instead of two.** Before this PR both Describes queued behind the stream on nearly every cycle; after it, at most one call can, and on an unloaded host neither does (~1.2-1.9ms, ten consecutive cycles). The PR improves the common case and cannot be worse than what it replaces — it just does not promise a guarantee.

Recorded the limitation in the code at the shared-describe construction site (40dd1ba) so the next reader does not have to rediscover it.

PR #209, opened 2026-09-18
Account the timer backfill in the units collection timings

cooperlees:
Fourth of the small remaining varlink items from the [#37](https://github.com/cooperlees/monitord/issues/37) audit: *"Time the timer backfill into inner `collection_timings`"*.

## Change

The varlink path backfills timer properties over D-Bus (`collect_all_timers_dbus`), but that phase was invisible in `collection_timings`: its duration was excluded from `per_unit_loop_ms` and `timer_dbus_fetches` stayed at 0. So the varlink path under-reported the D-Bus work it actually does, and the two paths' inner timings were not comparable.

`merge_timer_stats` now takes the phase duration and accounts it the way the D-Bus path accounts its own timer work — duration into `per_unit_loop_ms`, one `timer_dbus_fetches` per timer resolved. Applied on both the host (`lib.rs`) and container (`machines.rs`) paths.

## Verification

Live on Fedora Rawhide (systemd 262~rc3), same two timers tracked on both paths:

```
D-Bus:   timer_fetches=2 service_fetches=0 state_fetches=2 per_unit_loop_ms=1.18
varlink: timer_fetches=2 service_fetches=1 state_fetches=0 per_unit_loop_ms=1.73
```

`timer_dbus_fetches` was `0` on the varlink path before this change. `state_dbus_fetches` stays 0 there by design — time-in-state comes from the `StateChangeTimestamp` metric, not a D-Bus fetch — and the README parity note now says exactly that instead of lumping the timer backfill in with it.

- `cargo test`: 131 passed (two new tests: the backfill adds to an already-recorded parse loop rather than replacing it, and a no-timers run still accounts its `ListUnits` cost).
- Varlink integration test on Rawhide: no fallback to D-Bus, outputs identical across 59 keys.


cooperlees:
All three addressed in the follow-up commit.

**Dropped duration on the error path** — correct, and the case that mattered most: a backfill that fails by timing out burns `dbus_timeout` seconds, which is exactly the number you would want visible in `per_unit_loop_ms`. Both the host and container paths now measure around the call and account the duration in either arm, via a `record_backfill_duration` helper so the accounting rule stays in `timer.rs` next to `merge_timer_stats`. No fetch is counted for a call that produced nothing.

**Stale in-source comments** — also correct, and a good catch since they feed the generated docs. `units.rs` and `varlink_units.rs` both said only `service_dbus_fetches` could be nonzero on the varlink path. Both now describe the two D-Bus phases that contribute (oneshot type lookups and the timer backfill) and state that `state_dbus_fetches` is the only counter that stays zero there, matching the README.

- `cargo test`: 132 passed, including a new case asserting a 5s failed backfill lands in `per_unit_loop_ms` with `timer_dbus_fetches` still 0.
- Varlink integration test on Rawhide: no fallback to D-Bus, outputs identical across 59 keys.

Issue #211, opened 2026-09-18
Container varlink collection cannot work from the host: systemd rejects cross-PID-namespace peers


cooperlees:
monitord collects container stats by reaching the container's sockets through `/proc/<leader_pid>/root/...`. That works for D-Bus. **It cannot work for varlink**, so every container varlink collector fails and silently falls back, on every collection cycle.

Found while building a systemd-nspawn fixture for the varlink integration test — the fixture that #37 lists as a prerequisite for machine enumeration. It failed on its first run.

## Evidence

`varlinkctl` against the container's own socket, so this is independent of monitord. Fedora Rawhide, systemd 262~rc3, nspawn machine `mtest`:

```
inner socket via /proc, PID ns entered : {"context":{"ShowStatus":true,...   works
inner socket via /proc, no nsenter     : Connection reset by peer            fails
outer socket, no nsenter (control)     : {"context":{"ShowStatus":true,...   works
```

The path resolves and the connection is accepted, then closed. Nothing is logged in the container's journal. Entering the container's PID namespace is sufficient to make the identical call succeed, which isolates the *trigger*: it is not the mount namespace, not the socket path, and not user namespacing.

**The mechanism is not confirmed.** A review of `varlink-unit.c`, `varlink.c` and `sd-varlink.c` found no explicit cross-PID-namespace check, so this likely falls out of peer pidfd/credential validation rather than a deliberate refusal — but it has not been pinned to a code site. The behaviour above is observed; the cause is inferred, and worth confirming upstream before quoting a reason.

D-Bus has no equivalent peer check, which is why container D-Bus collection works today.

## What this means in practice

With a real machine present, every container varlink path falls back:

```
Varlink networkd stats failed for container mtest, falling back to file-based
Varlink system state failed for container mtest, falling back to D-Bus
Varlink version failed for container mtest, falling back to D-Bus
Varlink units stats failed, falling back to D-Bus: Connection reset by peer (os error 104)
```

So the container varlink code added in #201 (timer backfill), #202 (`time_in_state_usecs`), #206 (version/system state) and the original container units path has never done anything. No data is lost — the D-Bus fallback produces correct output — but:

- Each container costs 4 guaranteed-to-fail varlink connects per cycle before doing the D-Bus work anyway.
- Daemon mode logs 4 warnings per container per cycle, which is pure noise.
- It blocks `io.systemd.Machine.List` adoption: enumerating machines over varlink is pointless while the per-machine collection underneath has to be D-Bus.
[...]

PR #212, opened 2026-09-18
Collect full per-service stats over varlink

cooperlees:
First of the `io.systemd.Unit.List` items from the [#37](https://github.com/cooperlees/monitord/issues/37) audit. Also finishes what #183 started.

## The gap

Services in `[services]` reported **0 of 16 populated fields** on the varlink path. The metrics endpoint exposes only `NRestarts` and `StatusErrno` per service, so CPU, memory, tasks, timestamps and timeouts all sat at zero while D-Bus returned real values:

```
D-Bus:   12/16 fields populated
varlink:  0/16 fields populated
```

#183 fixed the *other* half of this (varlink minting a zeroed entry for every service on the box); the residual was that listed services still reported untrue zeros.

## Change

New `io.systemd.Unit.List` client (systemd v258+) filling `ServiceStats` from `context.Service` + `runtime.CGroup`/`runtime.Service`. The service type comes from the same reply, which **retires the D-Bus oneshot lookup** — the varlink path no longer reaches back to D-Bus for it, except as a fallback if the manager socket is unusable.

**Per-unit calls, not the streaming form.** Measured on a 305-unit host:

```
bulk Unit.List (--more):  92.0ms   1.3MB, all 305 units
per-unit Unit.List:        0.47ms  mean on a warm connection
break-even:              ~198 units
```

monitord wants a handful — the `[services]` list plus oneshot candidates — so streaming would turn a 16ms cycle into a 110ms one. Replies are cached for the collection run, so a unit wanted by two consumers costs one call. Observable at debug level:

```
Varlink unit details: 3 unit(s) fetched for 2 service(s) + 1 oneshot candidate(s) in 1.40ms
```

`processes` comes from `cgroup.procs` under the path in the reply, since systemd exposes no per-cgroup process count over varlink (noted in #37).

## The defaults trap

systemd omits fields sitting at their default rather than sending zero, so each mapping has to restate what the D-Bus property returns for the same unit — `u64::MAX` for unset IO counters and `TimeoutCleanUSec` (`infinity`), `0` for `WatchdogUSec`. Getting this wrong would silently diverge only on hosts with IO accounting off, which is most of them. Unit tests assert each default.

One field cannot reach parity: `active_exit_timestamp` stays 0, because systemd declares `ActiveExitTimestamp` in the IDL but never emits it (already tracked in #37).

## Verification

Live on Fedora Rawhide (systemd 262~rc3), same service on both paths — all 16 fields match, with only the live-sampled counters differing between runs seconds apart:

```
                          D-Bus                  varlink
active_enter_timestamp    1789701442296989       1789701442296989
inactive_exit_timestamp   1789701442287084       1789701442287084
state_change_timestamp    1789703841346990       1789703841346990
ioread_bytes              18446744073709551615   18446744073709551615
timeout_clean_usec        18446744073709551615   18446744073709551615
processes                 2                      2
tasks_current             2                      2
restart_usec              100000                 100000
cpuusage_nsec             103254000              103715000   (live)
memory_current            4227072                4157440     (live)
```

The integration test now points `[services]` at a real unit, so this is enforced in CI: **79 keys compared, up from 59**, with only the three live cgroup counters excluded. `cargo test`: 135 passed; Python helpers: 15 passed. `cargo fmt --check` and `cargo clippy` clean.


cooperlees:
Six of the eight were real bugs — thanks, these were good catches. Verified each against systemd 262 before changing anything, which turned up two corrections to the review itself and one to #37.

## Fixed

**`TimeoutCleanUSec` under `context.Exec`** — correct. Confirmed with a service carrying `TimeoutCleanSec=30s`:

```
context.Service: {TimeoutStopUSec, TimeoutStartFailureMode, ..., WatchdogUSec}
context.Exec:    {TimeoutCleanUSec: 30000000}
```

It only looked right before because the default is infinity, which matches the fallback. Now reads 30000000 on both paths.

**`ActiveExitTimestamp` is emitted** — correct, and this also corrects #37, which recorded it as never sent. That was a sampling artifact: it is absent only until a unit has actually left the active state. `systemd-networkd.service` has one, and both paths now agree:

```
dbus    active_exit_timestamp = 1789702478285190
varlink active_exit_timestamp = 1789702478285190   (was hardcoded 0)
```

**Omitted cgroup counters must be `u64::MAX`** — correct, and the most damaging of the eight. With accounting off, zero claims a service uses no memory rather than that it is unmeasured:

```
noacct.service (MemoryAccounting=no, TasksAccounting=no)
  dbus    memory_current=18446744073709551615  tasks_current=18446744073709551615
  varlink memory_current=18446744073709551615  tasks_current=18446744073709551615
```

**`get()` swallowing transport errors** — correct. It returns `Result` now, so a broken socket reaches the caller and triggers the fallback instead of looking like `NoSuchUnit` and reporting success with empty stats.

**Process count vs `GetProcesses`** — correct. Now walks the cgroup subtree and folds in `MainPID`, since systemd does both.

**Incomplete fallback** — correct. A failed detail pass now redoes the whole units collection over D-Bus; restoring only the oneshot override left `[services]` half filled from the metrics, which is worse than either path alone.

## Corrected

**Minimum version: v261, not v258 and not v260.** I had v258 (wrong), you said v260 (also wrong). `Unit.List` exists in v258, but the context sections monitord reads come from `src/core/varlink-service.c` and `src/core/varlink-timer.c`, which first appear in v261:

```
varlink-service.c  v259 ABSENT  v260 ABSENT  v261 present
varlink-timer.c    v259 ABSENT  v260 ABSENT  v261 present
```

README now says v261+ and explains why.

## Not doing

**Wiring `apply_unit_details` into the container path.** That call cannot work: systemd refuses varlink peers from another PID namespace, so a container's unit socket is unreachable through `/proc/<leader>/root/...`. Confirmed with `varlinkctl`, independently of monitord, in #211 — the connection is accepted and then reset. Containers deliberately keep the full D-Bus path, which works. Adding the call would only add a guaranteed failure and a warning per container per cycle.

## Two things you did not flag

- **`WatchdogUSec` differs in kind.** varlink gives the *configured* value; the D-Bus property is the *currently armed* one, which reads `infinity` while a service is not running. They agree for running services, and systemd exposes no runtime equivalent over varlink. Documented in the type.
- **A `[services]` unit that systemd has GC'd out of `ListUnits` is reported by varlink but not D-Bus** (`LoadState=loaded`, absent from `list-units`). `Unit.List` loads by name; `ListUnits` only returns what is currently listed. Arguably varlink is more useful here — the user did ask for that service — but it is a real behavioural difference, so flagging it rather than burying it.

`cargo test`: 135 passed. Integration test on Rawhide: no fallback, 79 keys identical, plus the accounting-off and `TimeoutCleanSec` cases checked by hand above.

cooperlees:
Follow-up, prompted by @cooperlees asking whether CI on Rawhide already covers v261.

It does — Rawhide gives us 262~rc3, so every v261+ path in this PR is exercised in CI. But that question exposed a real bug in what happens *below* v261, which CI can never reach.

**v258–v260 were silently wrong.** `Unit.List` exists on those versions and answers **successfully** — it just returns no per-type context sections, because `src/core/varlink-service.c` only lands in v261. Confirmed in the v260 source: `unit_context_build_json()` is there and builds generic unit fields (`Wants`, `Requires`, …) with no `Service` section.

So on a v258–v260 host this PR would have mapped every field to its default, treated the call as a success, and **never fallen back to D-Bus** — reporting `u64::MAX` memory and `0` restart delays as though they were real, instead of failing visibly.

A `.service` reply carrying no `context.Service` is now an error, which sends the whole units collection back to D-Bus. A non-service name in `[services]` is a config mistake rather than an old systemd, so it does not trip the check.

Pinned by a unit test, since no CI job can run a v260 host — the whole point is that this shape is unreachable on Rawhide.

`cargo test`: 136 passed. Integration test still green on Rawhide: no fallback, 79 keys identical.

PR #213, opened 2026-09-18
Collect timer stats over varlink instead of the D-Bus backfill

cooperlees:
Second of the `io.systemd.Unit.List` items from [#37](https://github.com/cooperlees/monitord/issues/37). Rebased onto main now that #212 has merged.

## Change

Timer properties were the last thing the varlink units collector reached back to D-Bus for: after every varlink collection it ran a full `ListUnits` plus per-timer property fetches. `context.Timer` and `runtime.Timer` carry all eleven `TimerStats` fields, and the triggered unit's state-change timestamps come from that unit's own reply.

**Enumeration costs nothing extra.** Timers are picked out of the metrics stream already fetched each cycle — every unit appears there as a metric object — rather than a second bulk call. Same gating as the D-Bus path (`[timers]` enabled, allowlist, blocklist).

**The cache now pays.** A timer's triggered unit is frequently also in `[services]`, and is fetched once instead of twice. Measured on the Rawhide fixture:

```
2 services + 1 oneshot candidate + 2 timers + 2 triggered units : 7 fetches
2 services + 2 timers + 2 triggered, one service *is* a target  : 5 fetches (6 distinct, 1 deduped)
```

The D-Bus backfill remains for when the unit socket is unusable, and for containers, which cannot reach their own varlink sockets (#211).

## Requires systemd v261+

`Unit.List` exists from v258, but `src/core/varlink-timer.c` — which emits the `Timer` context — first appears in v261, exactly like `varlink-service.c`. On v258–v260 the call succeeds and returns no `Timer` section, so `require_unit_context` rejects such a reply and the whole units collection is redone over D-Bus rather than reporting zeroes as though they were measurements.

## Verification

Live on Fedora Rawhide (systemd 262~rc3) — all 22 timer fields across both timers are **byte-identical** between the paths (11 fields × 2 timers), including the sentinel-ish ones:

```
                                       D-Bus                 varlink
dnf-makecache.timer
  accuracy_usec                        60000000              60000000
  last_trigger_usec                    1789702499149363      1789702499149363
  next_elapse_usec_monotonic           15781381588           15781381588
  next_elapse_usec_realtime            0                     0
  randomized_delay_usec                3600000000            3600000000
  service_unit_last_state_change_usec  1789702499313910      1789702499313910
systemd-tmpfiles-clean.timer
  randomized_delay_usec                0                     0            (omitted by systemd)
  remain_after_elapse                  1                     1
timer_persistent_units                 0                     0
timer_remain_after_elapse              2                     2
```

The integration test's timers allowlist is already emptied, so every timer on the host is compared in CI.

**The varlink units collector now makes no D-Bus calls at all:**

```
varlink dbus fetch counters: timer=0 service=0 state=0
```

## Known difference

If a timer's triggered unit cannot be resolved, varlink keeps the timer with its own fields intact and the two trigger timestamps at 0; the D-Bus path drops the timer entirely, because its `GetUnit` call errors and the whole per-timer fetch is abandoned. Keeping it is the more useful behaviour — a timer whose target is unloaded still has a real accuracy, next elapse and last trigger — but it is a genuine difference, recorded in the code rather than buried.

- `cargo test`: 139 passed (timer mapping, the omitted-`RandomizedDelayUSec` default, an unresolvable trigger target, and the pre-v261 guard for both sections).
- Integration test: no fallback to D-Bus, 79 keys identical.


cooperlees:
Rebased onto main now that #212 has merged — this is a single commit against main again, no longer stacked.

The rebase was not mechanical: #212's review changed the types this builds on, so the timer code had to be reconciled with them.

- `ServiceContext` lost `timeout_clean_usec` (it moved to the new `ExecContext`) and `UnitRuntime` gained `active_exit_timestamp`, so every timer test fixture needed updating.
- `UnitLookup::get` now returns `Result`, so the timer and triggered-unit lookups propagate transport errors rather than treating them as "no such unit".
- The two fallbacks collapsed into one. This PR previously restored timers and the oneshot override separately on failure; #212 made a failed detail pass redo the whole units collection over D-Bus, which covers timers too. The piecemeal restore is gone.

Also corrected the version claim here the same way: **v261+**, not v258. `varlink-timer.c` first appears in v261, exactly like `varlink-service.c`.

Re-verified after the rebase on Rawhide (systemd 262~rc3): 139 tests pass, integration test green with no fallback and 79 keys identical, and the 22 timer keys hash identically across both paths:

```
monitord-ci:          22 timer keys, digest=591897ba637c
monitord-varlink-ci:  22 timer keys, digest=591897ba637c
```

Call count unchanged: `7 unit(s) fetched for 2 service(s) + 1 oneshot candidate(s)` — 2 services + 2 timers + 2 triggered units + 1 oneshot candidate, deduped by the per-run cache.

cooperlees (commented):
**Muse code review** (posted for @cooperlees — he is out of Copilot review tokens, so I did the systemd API/version audit against a local rebased systemd checkout).

I verified every systemd claim in this PR against upstream. Most check out with commit-level evidence. One real issue, plus a few nits.

## Verified ✅

**v261 minimum for `TimerContext`/`TimerRuntime`** — confirmed. `ae619b1224` ("core: implement TimerContext/Runtime for io.systemd.Unit.List + tests", 2026-05-07) first appears in tag `v261` (`git tag --contains` → v261, v261-rc1… first). The README wording ("`varlink-service.c` and `varlink-timer.c` first appear in v261") is accurate for the sections monitord reads. For completeness: `Unit.List` skeleton itself is v258 (`472abf7bec`), `ExecContext` is v259 (`fdb2c0dd6f`) — so the "method exists in v258, sections land in v261" framing is exactly right.

**Omitted-`RandomizedDelayUSec` → 0** — confirmed. `c5406d3e1e` (also v261) switched it to `JSON_BUILD_PAIR_FINITE_USEC_NON_ZERO`, which omits zero (`sd-json.c`: skipped unless `u != USEC_INFINITY && u > 0`). D-Bus side is `bus_property_get_usec` on the raw field, so unset reads 0 there. `unwrap_or(0)` matches. Same logic covers `LastTriggerUSec` (`JSON_BUILD_PAIR_DUAL_TIMESTAMP_NON_NULL` omitted unless `dual_timestamp_is_set`; D-Bus `BUS_PROPERTY_DUAL_TIMESTAMP` always emits raw offsets, 0/0 when never fired). ✅

**Monotonic-only timer reports realtime 0 on both paths** — confirmed. `timer.c` inits `next_elapse_realtime = 0` and only fills it for calendar triggers, so D-Bus reads 0 and varlink `FINITE_USEC` emits 0 (0 ≠ `USEC_INFINITY`, and it's not a `NON_ZERO` field). The test comment is correct.

**Empty trigger name → error! + zero timestamps, both paths** — confirmed. `bus_property_get_triggered_unit` returns `""` when no trigger; varlink uses `STRING_NON_EMPTY` so the key is absent. Both log the same `"No service unit name found for timer."` string. ✅

**"Every unit appears in the metrics stream"** — confirmed. `unit_active_state_build_json` / `unit_load_state_build_json` in `varlink-metrics.c` iterate all of `manager->units` (aliases skipped, same as `ListUnits`), emitting per-unit objects, so deriving timer names from `metric.object()` mirrors the D-Bus `ListUnits` scan.

**"All 22 timer fields byte-identical"** — arithmetic checks out: 11 `TimerStats` fields × 2 timers.

## Issue 1 (should fix): no pre-v261 guard on the timer path

`require_service_context` exists precisely because v258–v260 answer `Unit.List` successfully but without the per-type sections, which would otherwise produce silently-wrong defaults instead of a D-Bus fallback. But only the `[services]` and oneshot loops call it. The new timer loop does not.

Consequence: on a **v260** host (metrics exist there, so the varlink path runs) with timers enabled but `[services]` empty and no oneshot candidates, every tracked timer maps to all-zero `TimerStats` — `context.Timer` is absent pre-`ae619b`, so `timer_triggered_unit` returns `None`, the `error!` fires, and zeros are inserted with real-looking success. No fallback ever triggers. This is exactly the failure class the service guard was added for. Suggested fix: a `require_timer_context`-style check in the timer loop that bails (→ whole-phase D-Bus redo), mirroring the services behavior.

## Issue 2 (minor): stale `parse_metrics` comment

`src/varlink_units.rs` (~L429–436, untouched by this diff) still says "Two later D-Bus phases add their own duration to `per_unit_loop_ms`… the oneshot type lookups (`service_dbus_fetches`) and the timer backfill (`timer_dbus_fetches`)". After #212 + this PR, neither fires on the host path — this comment now describes the container path only. Worth a one-hunk refresh in this PR since it completes the removal.

## Nits / questions

- **Fetch arithmetic, PR body vs commit message:** body says "2 services + 2 timers + 2 triggered units, no overlap: **7** fetches"; the commit message says "**6**". 2+2+2 = 6 distinct cache entries, and the overlap case (5) agrees with 6. The body looks like the typo.
- **`GetUnit`-failure divergence (edge case):** if a timer's `Unit=` names a unit that isn't loaded, D-Bus `GetUnit` errors with `?` and the timer is *dropped* from output; varlink caches `None` and *keeps* the timer with zeroed trigger timestamps. Same as empty-name only when the name is empty. Rare (trigger targets are normally loaded), but the "byte-identical" claim has this one hole — consider a test or a `continue` to match D-Bus.
- **Container PID-namespace wording:** #211's evidence is empirical (`Connection reset by peer`, fixed by entering the PID ns). I grepped `varlink-unit.c`, `varlink.c`, and `sd-varlink.c` and found no explicit cross-PID-namespace check — the refusal likely falls out of peer-pidfd/credential validation, but I couldn't pin it to a code site. The behavior claim is fine (it's observed), but "systemd refuses varlink peers from another PID namespace" states a mechanism; consider softening to "a container's unit socket is unreachable from the host (see #211)" unless upstream confirms the mechanism.
- **Drive-by, out of scope:** D-Bus timer enumeration uses `.contains(".timer")`, varlink uses `.ends_with(".timer")`. The latter is strictly more correct (a hypothetical `x.timer.service` would be misclassified by the former). Not this PR's problem, but now the two paths differ textually.

Nothing above disputes the core change: the field mapping is faithful to `varlink-timer.c`, the version floor is right, and the cache-sharing is sound. Issue 1 is the only one I'd want closed before merge.

cooperlees:
Thanks — this is a genuinely better review than I got from Copilot, and **Issue 1 was a real bug**. Fixed in d2e97f4, along with everything else.

## Issue 1 — pre-v261 guard missing on the timer path

Correct, and embarrassing: I added `require_service_context` in #212 for exactly this failure class and then didn't extend it to the loop this PR introduces. Your v260 scenario is precisely right — timers enabled, `[services]` empty, no oneshot candidates, and every timer maps to all-zero `TimerStats` while reporting success, so nothing ever falls back.

It is now one `require_unit_context` covering both sections, called from all three loops (`[services]`, oneshot candidates, timers). A unit type monitord does not map stays exempt, since that is a config mistake rather than an old systemd. The test now asserts the timer case explicitly — my previous test actively asserted the *wrong* thing (`require_service_context("foo.timer", …).is_ok()`), which is how the gap survived.

## Issue 2 — stale `parse_metrics` comment

Fixed, and you were right that this PR is the place: it completes the removal. It now says all three `*_dbus_fetches` stay 0 on the host path, and that they only move for containers or when the unit socket is unusable.

## Nits

**Fetch arithmetic** — you're right that the body didn't add up, though the measured `7` was correct: the breakdown omitted the oneshot candidate. It is 2 services + **1 oneshot candidate** + 2 timers + 2 triggered = 7 distinct. The commit message's `6` describes the *other* config (2 services + 2 timers + 2 targets, no oneshot candidates), where the overlap case gives 5. PR body corrected to state both configs explicitly.

**`GetUnit`-failure divergence** — confirmed by reading `timer.rs`: `get_unit(...).await?` propagates, `collect_all_timers_dbus` logs and skips, so the timer is dropped from output entirely. varlink keeps it with zeroed trigger timestamps. I've kept the varlink behaviour and documented it rather than adding a `continue`: a timer whose target is unloaded still has a real accuracy, next elapse and last trigger, and dropping all of that to match a D-Bus limitation seemed the wrong trade. It is now called out in the code and in the PR body under "Known difference", so the byte-identical claim is properly bounded. Happy to switch to `continue` if you'd rather have strict parity.

**PID-namespace wording** — fair, and I over-claimed. I observed the behaviour (accepted then reset; succeeds from inside the namespace) but never pinned it to a code site either. The code comment now says the connection "is accepted and then reset unless the caller is inside the container's PID namespace (observed)", and I'll soften #211 the same way.

**`.contains(".timer")` vs `.ends_with(".timer")`** — good spot, and agreed the D-Bus side is the loose one. Out of scope here; I'd rather fix `timer.rs` in its own change than alter D-Bus behaviour inside a varlink PR. Happy to send that separately.

Re-verified after the fixes: 139 tests pass, integration test green on Rawhide with no fallback and 79 keys identical.

PR #214, opened 2026-09-18
Match timer units by suffix rather than substring

cooperlees:
Raised by Muse while reviewing #213: the varlink timer enumeration added there matches `.ends_with(".timer")`, while the existing D-Bus paths use `.contains(".timer")`. The latter is the loose one.

## The bug

Both D-Bus paths pick timers out of a list of *every* unit on the system, so a substring match also selects units that merely mention a timer in their name:

```
my.timer.service     contains ".timer" -> selected as a timer
foo.timers.service   contains ".timer" -> selected as a timer
```

Such a unit is then put through `collect_timer_stats`, whose `TimerProxy` call fails because the unit has no timer interface, and monitord logs `Failed to get my.timer.service stats: ...` for a unit that was never a timer.

There were **two** sites, not one — `timer.rs:190` in `collect_all_timers_dbus` and `units.rs:803` in the per-unit collection loop. Both now use `ends_with`, matching what both paths already do for `.service`, behind a `SYSTEMD_TIMER_SUFFIX` constant alongside the existing `SYSTEMD_SERVICE_SUFFIX`.

## Verification

New test pins the cases that motivated it, including the ones the old check got wrong:

```rust
assert!(selects("logrotate.timer"));
assert!(!selects("my.timer.service"));
assert!(!selects("foo.timers.service"));
assert!(!selects("timer.target"));
```

`cargo test`, `cargo fmt --check` and `cargo clippy` all clean. No behaviour change for any normally-named unit, so the varlink-vs-D-Bus parity run is unaffected.

Kept separate from #213 rather than folded in: that PR moves timers onto varlink, and this changes D-Bus behaviour, so they are easier to reason about apart. #213 adds the same constant, so whichever lands second needs a one-line rebase.


cooperlees (commented):
**Muse code review** (posted for @cooperlees).

Small but correct. This fixes the drive-by I noted while reviewing #213 (the `.contains(".timer")` vs `.ends_with(".timer")` mismatch — note for the record this is the suffix-matching nit, not Issue 1, the missing pre-v261 timer guard, which is still open on #213).

## Checked

- **Both sites found.** I grepped `main` for timer-name matching: exactly the two sites this PR touches (`timer.rs:190`, `units.rs:803`), no others in `src/`, `tests/`, or workflows. Nothing left behind.
- **The bug claim is accurate.** `units.rs:803` failure path confirms it: a mis-selected `my.timer.service` goes through `collect_timer_stats` → `TimerProxy` on a unit with no timer interface → `error!("Failed to get {} stats: ...")` for a non-timer. Real log noise on top of the wasted D-Bus round trips.
- **Suffix is the semantically right match, not just a tighter one.** systemd determines unit type from the name suffix (`.timer` → `UNIT_TIMER`), so `ends_with` aligns monitord with how systemd itself classifies units. Precedent in-repo agrees: `.service` matching already uses `ends_with(SYSTEMD_SERVICE_SUFFIX)` in both `units.rs` and `varlink_units.rs`. This PR just brings timers into line.
- **Test is honest.** It tests the predicate via the real constant rather than hardcoding `".timer"` — so it genuinely guards against regression to `.contains`, not just documents examples. The `foo.timers.service` case is a nice catch (substring match *without* an embedded `.timer.` boundary).
- **Rebase note is correct.** #213 adds the identical constant, so second-to-land needs a one-line rebase — and #213's `parse_metrics` timer enumeration already uses `ends_with(SYSTEMD_TIMER_SUFFIX)`, so after both land all three enumeration sites agree.

No issues. `ends_with` on a `&str` suffix constant, two call sites, one test — nothing to push back on. LGTM.

PR #215, opened 2026-09-18
Collect boot blame from the metrics stream instead of per-unit D-Bus

cooperlees:
Third of the `io.systemd.Unit.List` items from [#37](https://github.com/cooperlees/monitord/issues/37) — though it needs no `Unit.List` call at all.

## Change

Boot blame did one `ListUnits` plus **two property reads for every unit on the system** — several hundred D-Bus round trips, which is why it ships disabled by default. The per-unit `ActiveTimestamp` and `InactiveExitTimestamp` metrics carry the same two values, so the varlink path computes the identical ranking from one streamed `List`.

**No extra call per cycle.** `update_boot_blame_stats` already caches in memory and on disk keyed by boot id, so a daemon pays the stream once on its first cycle and reuses the result. D-Bus remains the fallback.

The one trap: `ActiveTimestamp` is emitted **twice** per unit, distinguished by an `event` field of `enter` or `exit`. Only `enter` corresponds to the `ActiveEnterTimestamp` property the D-Bus path read; picking up `exit` (always 0 for a running unit) would invert the calculation and silently produce nonsense. A test covers it.

## Verification

Live on Fedora Rawhide (systemd 262~rc3), caching disabled so each run really collects — byte-identical, same order:

```
blame-dbus:     {'dev-vda1.device': 369.201813, 'systemd-udev-trigger.service': 0.042779,
                 'systemd-networkd.service': 0.036911, 'systemd-nspawn@mtest.service': 0.028748,
                 'systemd-oomd.service': 0.025262}
blame-varlink:  {'dev-vda1.device': 369.201813, 'systemd-udev-trigger.service': 0.042779,
                 'systemd-networkd.service': 0.036911, 'systemd-nspawn@mtest.service': 0.028748,
                 'systemd-oomd.service': 0.025262}
```

And roughly **5x cheaper**, three runs each:

```
blame-dbus:     57.1ms  41.9ms  42.9ms
blame-varlink:   8.9ms   7.6ms   7.7ms
```

- `cargo test`: 143 passed — ranking, truncation to `num_slowest_units`, allowlist/blocklist, units that never activated, and the enter/exit event distinction.
- Integration test on Rawhide: no fallback to D-Bus, 79 keys identical.

Note the integration test does not cover this yet: `[boot]` ships disabled, so neither path collects it in CI. Enabling it there would make the comparison meaningful — happy to do that here or separately, though boot blame on the D-Bus side costs ~50ms per run.


cooperlees:
Added CI coverage for this, as asked — the parity run now compares boot blame instead of skipping it.

**79 keys → 84 keys**, the five being the boot blame entries:

```
{'dev-vda1.device': 369.201813, 'systemd-udev-trigger.service': 0.042779,
 'systemd-networkd.service': 0.036911, 'systemd-nspawn@mtest.service': 0.028748,
 'systemd-oomd.service': 0.025262}
```

**The cache had to go off, and that is the interesting part.** Both runs share `cache_dir`, so with caching left on the D-Bus run would have read back the cache the varlink run had just written. The outputs would have matched because they came from the *same* collection — green, and proving nothing. The generated CI config now sets `cache_enabled = false` so each run genuinely collects.

That change also broke an existing test in a useful way: `test_configs_differ_only_in_varlink_toggle` did a blanket replace of `enabled = false` → `enabled = true`, which now also rewrote `cache_enabled`. Rather than loosen it, it compares line by line and asserts the single differing line is the one inside `[varlink]` — stricter than before.

Stable across three consecutive runs (84 keys each time). Boot blame values are absolute boot-time measurements, so they do not drift between the two invocations seconds apart; the only way the set could shift is a unit activating in that window with an activation slower than the current fifth place (0.025s), which does not happen in a quiescent container.

Python helpers: 16 tests passing.

cooperlees (commented):
**Muse code review** (posted for @cooperlees — same arrangement as #213/#214: systemd API and version claims audited against a local rebased systemd checkout).

## Verified ✅

**The enter/exit trap is real, and the fix is right.** `active_timestamp_build_json` (`src/core/varlink-metrics.c`) emits `ActiveTimestamp` twice per unit — `unit->active_enter_timestamp.realtime` with `fields: {event: enter}` and `unit->active_exit_timestamp.realtime` with `{event: exit}`. The `event == "enter"` guard plus the test pinning it are exactly the right response; without it a running unit's `exit` value (always 0) would flow into `active_enter_usec` and `saturating_sub` would collapse the blame to 0. The D-Bus `BUS_PROPERTY_DUAL_TIMESTAMP("ActiveEnterTimestamp", …)` reads `active_enter_timestamp.realtime`, so both paths measure the same clock. ✅

**Version floor is right-ish.** `ActiveTimestamp`/`InactiveExitTimestamp` arrive in `4ec0520d9a` (2026-03-19, first in tag `v261`), while the metrics skeleton is `bb1ef2edf7` (v260) — so "v261+, which is when the metrics endpoint gained [these] families" is accurate. One wrinkle, same class as the Issue-1 I raised on #213: **there is no pre-v261 guard here.** On v260 the socket answers successfully but these families don't exist, so `collect_activations` yields `{}` and blame silently reports empty rather than falling back to D-Bus. Mitigating factors: blame ships disabled, and the D-Bus fallback *does* run on socket errors — but v260 is a silent-empty case, not a loud fallback. Worth either a `Describe`-based family check or at minimum a `warn!` when varlink is enabled, families are absent, and blame is requested. Not a merge-blocker given the feature is opt-in, but the same failure class as #213/I1.

**Arithmetic parity holds.** D-Bus: `active_enter.saturating_sub(inactive_exit) / 1e6`, dropping `0.0`. Varlink: same `saturating_sub`, same divisor, same `> 0.0` filter, plus the explicit either-timestamp-zero drop that matches `get_unit_activation_time`'s early `Ok(0.0)`. Sort comparator (`partial_cmp` descending, `Equal` fallback) is even copied verbatim. The claimed byte-identical output is credible.

**"No extra call per cycle" is accurate.** `update_boot_blame_stats` returns early on in-memory cache and on-disk cache keyed by boot id, before the varlink/D-Bus branch — the stream is paid once per boot. And the 5x speedup in the body is consistent with the removed work (hundreds of per-unit `UnitProxy` constructions → one stream).

## Issues

**1 (minor): silent drop of non-i64 values.** `if !metric.value().is_i64() { continue; }` — systemd sends these via `metric_build_send_unsigned` (`sd_json_variant_new_unsigned`, u64). Real usec timestamps (~1.7e15) fit i64 comfortably, so this never fires today. But it's a filter on a wire-type technicality rather than a value check, and a future large/odd value would vanish without a `debug!`. Suggest at least logging the skip, or accepting `is_u64() || is_i64()`.

**2 (nit): README version line.** The new "**Boot blame** (`io.systemd.Metrics` — systemd v261+)" entry is correct, but the adjacent "**Units** (`io.systemd.Metrics` — systemd v260+, v261+ for …)" line doesn't mention blame's families. A reader could reasonably infer all Metrics families follow the Units line's floor/Fallback rules. One clause ("plus `ActiveTimestamp`/`InactiveExitTimestamp` for boot blame") on the Units line would close it.

**3 (nit, integration test):** the `test_configs_differ_only_in_varlink_toggle` rewrite from blanket-replace to line-diff is good — the old version would indeed have rewritten `[boot] cache_enabled` too once blame was enabled. But it now asserts *exactly one* differing line, which makes the next collector that legitimately needs a CI-only config tweak (like this one) fail the test until the assertion is touched again. Consider asserting the differing set *contains* the varlink toggle rather than *equals* it. Pre-existing pattern, just now load-bearing.

Nothing here challenges the approach. The enter/exit distinction — the one thing in this PR that could have produced silently-wrong numbers — is handled and tested. LGTM with the v261-guard question answered or tracked.

PR #217, opened 2026-09-18
Enumerate verify units from the metrics stream instead of D-Bus

cooperlees:
Last primary-path D-Bus enumeration (see #37): `verify` called `ListUnits` before running `systemd-analyze`.

## Changes

- New `varlink_verify` module enumerates unit names from per-unit `UnitLoadState` metric objects (v260+ families, so no new minimum version). Verified live that the metric objects are exactly the set `ListUnits` returns (199 == 199, including `not-found` units) — an earlier revision wrongly dropped those and diverged by 6 units.
- `verify.rs` refactored so both paths share the allowlist/blocklist filter and the analyze+parse+count phase; varlink first with D-Bus fallback, mirroring `boot`.
- **Latent bug fix**: unit names starting with a dash (`-.mount`) were passed to analyze without `--`, so it exited with `unrecognized option '-.'` before checking anything — full verify always reported `total=0` on every system, on both paths. Added the separator plus a command-builder test.
- Integration test enables `[verify]` restricted to the fixture units; README capability bullet.

## Verification

- `cargo test`: 151 passed; integration-test unittests: 17 passed; `cargo fmt` clean.
- Live on Rawhide (systemd 262): full 199-unit verify agrees exactly on both paths (177 failures, all types match); fixture-restricted run also identical. (Full container test binary not runnable here — no container DNS for the in-container build — so the binary was host-built and the test's own config/compare functions were reused directly.)


cooperlees:
_Posted by Claude (Claude Code) acting on Cooper's behalf._

## Version claims: check out

The `systemd v260+` claim for `UnitLoadState` is consistent and correct. The metrics socket itself is the v260 floor (`README.md:755`), and the repo already documents that on v260 load-state counts are *"counted from per-unit load states"* (`README.md:776`) — i.e. the per-unit `UnitLoadState` family is exactly what's assumed present on v260. So this genuinely adds no new minimum, unlike boot blame's v261 timestamp families. The README bullet placement and the "falls back to D-Bus" wording match the existing entries.

Confirmed on the review host: systemd 259, and `/run/systemd/report/` does not exist — consistent with the v260 floor.

## The `--` bug fix is real

Verified live:

```console
$ systemd-analyze verify -.mount
systemd-analyze: invalid option -- '.'
systemd-analyze: invalid option -- 'o'

$ systemd-analyze verify -- -.mount -.slice dbus-broker.service
/etc/systemd/system/metatax.slice:4: Unknown section 'Service'. Ignoring.
```

Tracing the old failure stderr through `parse_verify_output` also confirms the `total=0` claim: the only dot-bearing token is `'.'`, whose first char fails `get_unit_type`'s alphanumeric/`-`/`\` check, so nothing was ever counted. Good catch, and the command-builder test pins it.

## Findings

**1. `varlink_integration_test.py:203` — the allowlist injection makes the parity test nearly vacuous.** Restricting both paths to `dbus-broker.service` + `kmod-static-nodes.service` means the comparison only proves those two names appear in both enumerations. Any divergence in the other ~197 units — the exact thing this PR is about, and the thing an earlier revision got wrong by 6 units — is filtered away before analyze runs. The runtime concern is legitimate, but this test can't catch the regression it's named for. Consider instead leaving the allowlist empty and diffing the enumerated *name sets* (e.g. via a debug log line, or output already in the comparison), rather than paying for a full analyze run.

**2. `varlink_integration_test.py:203` — leftover section lines.** The `[verify.allowlist]` branch appends the header plus two units then `continue`s, but only for the header line. The original section body (`# example.service`, `# example.timer`) still falls through the `elif` chain and gets appended after. Harmless today since they're comments, but if anyone ever uncomments an entry in `monitord.conf` the test silently starts verifying extra units. The `[timers.allowlist]` branch above handles this by skipping body lines explicitly.

**3. `src/varlink_verify.rs:47` — silent shrinkage on non-string load states.** `has_load_state_metrics` guards only that the *family* exists; `collect_unit_names` then drops any object whose value isn't a string. A partially malformed stream yields a smaller-than-`ListUnits` set, `verify_units` happily verifies the subset, and the result is reported as a successful full verify with no fallback. Given the module's whole premise is set-equality with `ListUnits`, worth either counting the skips and bailing if any occurred, or at least `warn!`ing rather than `debug!`ing.

**4. `src/varlink_verify.rs:62` / `src/verify.rs:206` — empty enumeration isn't distinguished from "nothing configured."** If the family is present but every object is skipped, `list_unit_names` returns `Ok(vec![])`, `verify_units` early-returns `VerifyStats::default()`, and output shows `total=0` — the exact symptom of the bug this PR just fixed, now un-fallback-able. A `bail!` on an empty name list (a running systemd always has units — the same reasoning the existing guard already uses) closes this.

**5. `src/varlink_verify.rs:36,52` — two full passes over the metric vector.** `has_load_state_metrics` scans, then `collect_unit_names` scans again. Minor, but folding the guard into the collect (return the names plus a `saw_family` flag, or just check `names.is_empty()` per #4) drops a pass and removes the possibility of the two predicates disagreeing.

## Checked and fine

- `update_verify_stats` is only called from `lib.rs:461`; `machines.rs` never collects verify, so the host-only `METRICS_SOCKET_PATH` constant is correct and containers are unaffected.
- The varlink/D-Bus dispatch shape matches `boot.rs:270` exactly — same `if config.varlink.enabled` / `match` / `tracing::warn!` fallback idiom.
- Analyze failures on the varlink path correctly propagate rather than falling back to D-Bus; only enumeration failures fall back. That's the right split.
- `filter_unit_names` is genuinely shared, so allowlist/blocklist semantics can't drift between paths.
- Sorting/dedup in the varlink path only changes analyze's argument order; `parse_verify_output` returns a `HashSet`, so stats are order-independent.
- `--` is handled by `getopt_long` in every systemd version, so the fix carries no version risk.

Nothing here is a blocker except arguably #1, which is a test-value issue rather than a correctness one.

PR #218, opened 2026-09-18
Per-section varlink toggles + show them in example monitord.conf

cooperlees:
## What
Each varlink-capable collector gets its own `varlink` toggle (default `true`), ANDed with the global `[varlink] enabled` master switch, so collectors can move to varlink one at a time instead of all-or-nothing:

- `[units]`, `[networkd]`, `[system-state]`, `[boot]` gate their host collectors
- `[machines] varlink` is an additional gate on all container varlink paths (containers may run older systemd without the varlink APIs)
- `[system-state] varlink` also gates the always-on version collection, which shares the host `Manager.Describe` call

The example `monitord.conf` now shows every toggle at its default with comments explaining the two-level scheme, and the README Enabling Varlink section documents it.

## Behavior
Defaults preserve current behavior exactly: global off (default) means all D-Bus/files as before; flipping the global on keeps every collector on varlink unless individually opted out. All varlink paths retain their existing D-Bus/file fallbacks.

## Tests
- Extended `test_full_config` (parses `varlink = false` in all five sections)
- New `test_per_section_varlink_toggle_defaults_and_override` (mixed opt-out)
- Default-true asserts in `test_default_config`
- `cargo test`: 147 passed; `cargo fmt --check` clean (clippy not installed locally, leaving to CI)

## Follow-ups (not in scope)
- `[verify]` gets no toggle here; it lands with the #217 rebase after merge.
- README's embedded config sample left as-is (trimmed illustrative copy, already omits whole sections).


cooperlees:
_Posted by Claude (Claude Code) acting on Cooper's behalf._

Reviewed the config changes. The core mechanism is sound: `read_config_optional_bool` returns `None` for a missing key (so existing configs keep working untouched) while still propagating `InvalidValue` for a malformed *value*, and every new field defaults `true`, so the "defaults preserve current behavior exactly" claim holds. The gates are consistently `global && section`, and the container paths correctly triple-AND with `[machines]`.

Findings below, roughly in order of how much I'd want them addressed.

## 1. `[system-state] varlink` governs version collection, which isn't in `[system-state]`

`src/lib.rs:254` and `src/machines.rs:229` gate `manager_describe` on `config.system_state.varlink`, and that value feeds the version collector — which is explicitly unconditional (`// Always collect systemd version`, `src/lib.rs:260`).

So a user who sets `[system-state] enabled = false` reasonably expects nothing else in that section to matter, but `varlink = false` there still flips version collection to D-Bus. The sharing of `Manager.Describe` makes this the natural implementation, and it *is* documented — but the knob lives in a section that doesn't govern the thing it affects. Worth either hoisting it (e.g. a `[monitord]`-level toggle) or making the docs explicit that this one key keeps acting even when its section is disabled.

## 2. `[timers]` is varlink-capable but gets no toggle, and silently follows `[units]`

Timers have no collector spawn of their own in `lib.rs` — they ride entirely on the units varlink path (`Unit.List` plus the `timer.rs` backfill), so `[units] varlink` is what actually controls their transport. The PR says "each varlink-capable collector gets its own `varlink` toggle", but `README.md:785` documents per-timer stats as a varlink-collected metric and it gets none.

Not a bug, but `monitord.conf` / README should state that timers follow `[units] varlink`. Otherwise someone will add `[timers] varlink = false` and it will parse clean and do nothing — see #3.

## 3. Unknown and misspelled keys are silently ignored

Because the new keys are read with `read_config_optional_bool`, a missing key is indistinguishable from an absent one. `[timers] varlink = false` (#2), or a typo like `varlnik = false` in any section, parses successfully and has no effect and no warning.

This is pre-existing behavior for optional keys, but this PR adds five new optional booleans across five sections and documents the pattern prominently, which makes the failure mode materially more likely to be hit. Worth considering a warn on unrecognized keys within known sections.

## 4. README understates the container gating as a two-way AND

`machines.rs` ANDs in the *host* section toggles for container collection: `networkd.varlink` at `:192`, `system_state.varlink` at `:229`, `units.varlink` at `:284`. So container collection is a three-way AND.

The `monitord.conf` comment gets this right ("Container collection additionally requires `[machines] varlink = true`"), but `README.md:771` only says a collector uses varlink "when both the global switch and its section toggle are true", which reads as two-way. Add the container case there.

## 5. Test gap: the AND semantics themselves are untested

`test_per_section_varlink_toggle_defaults_and_override` covers one opt-out under a global-on, and `test_full_config` covers all-false. Neither covers:

- an explicit `varlink = true` round-tripping (only the default and `false` are exercised, so a parse that ignored the key entirely would still pass the default asserts)
- `[varlink] enabled = false` with a section toggle `true` staying off

The second matters most: the conjunction is the entire point of the PR, and it lives in `lib.rs`/`machines.rs` rather than `config.rs`, so nothing currently tests it at any level. Even a small helper (`fn use_varlink(global: bool, section: bool) -> bool`) that the call sites share would make it unit-testable and remove the five hand-written `&&` chains that have to stay in sync.

## 6. Minor: comment duplication in `monitord.conf`

The same "Follows `[varlink] enabled`; set false to keep this collector off varlink" comment is repeated above four sections, plus the longer block before `[varlink]` that already explains the whole scheme. The block is enough; the per-section repeats could shrink or drop.

---

Nothing here blocks correctness — defaults genuinely preserve existing behavior. #1 and #2 are the ones I'd want resolved before merge, and both are arguably documentation fixes rather than code changes.

PR #219, opened 2026-09-18
Track per-collector varlink usage for Grafana adoption reporting

cooperlees:
## What

Each varlink-capable collector records which transport served it on the last run, exposed as `varlink_usage.<collector>` 0/1 gauges in the existing output formats (1 = varlink, 0 = D-Bus or the file-based networkd fallback). No new output plumbing, no exporter code, no new dependencies — the gauges ride the existing `json.rs` flattening, and downstream consumers such as monitord-exporter aggregate them from there.

Disabled collectors and collectors with no varlink path (`pid1`, `dbus_stats`) emit no gauge, so the present gauges are exactly the enabled set — no separate enabled-collectors counter needed. As the systemd under monitord upgrades past each endpoint's minimum version, collectors silently flip from 0 to 1 with no config change, so watching these gauges shows varlink adoption climbing across the fleet.

## Changes

- New `CollectorTransport` (`#[repr(u8)]`, `Serialize_repr`/`Deserialize_repr`, `Dbus=0`/`Varlink=1`) + `VarlinkUsage` types in `lib.rs`, recorded at each success/fallback point (host collectors in `lib.rs`, container paths in `machines.rs`, plus `boot.rs`/`verify.rs`); flattened in `json.rs` for host + per-machine stats. Same integers in json, json-pretty and json-flat
- Boot-blame disk cache stamps the producing transport; hits replay it so the gauge set stays stable run to run. Legacy files are treated as a miss, re-collected once, and rewritten in the new format
- `varlink_integration_test.py` gains `assert_varlink_usage`: checks the exact gauge set and values on both runs (D-Bus run all 0s, varlink run all 1s except host `machines` enumeration). Kept alongside the existing log-scrape fallback check as a positive assertion
- Documents the measured networkd varlink cost (~15ms vs files, daemon-side serialization) on `get_networkd_state`
- README: new Varlink usage gauges section + sample json-flat gauges; AGENTS.md: new-output-values-go-in-README convention

## Edge cases (for reviewers)

- `Dbus` also covers the file-based networkd fallback: the gauge answers "did this collector use varlink", not "which fallback served it"
- In-memory boot-blame hits keep the gauge the collecting run recorded; disk hits replay the stamped transport
- A container `units` gauge of 1 still includes D-Bus calls underneath (timer backfill + oneshot override, see #211) — compare host and container gauges separately
- Host `machines` gauge covers enumeration only, stays 0 by design (machined has no varlink List API, see #37)
- Full detail + test evidence in the commit messages (`git show` / `git log`)

## Test evidence

- `cargo test`: 158 passed; `./varlink_integration_unittests.py`: 29 passed; fmt clean, no new clippy warnings
- Live systemd 259 box: D-Bus run 0/5 on varlink, varlink run 3/5 (60%) — outputs otherwise identical; boot-cache replay and legacy-file self-heal verified live
- Live systemd 262 (Rawhide container, `./varlink_integration_test.py`): varlink run 6/7 (86%), parity holds

Review history: Claude Code (Opus 5) reviewed 9b092e6 — findings 1–3 + smaller items all addressed in 2ee2a69 (see that commit message). Copilot could not review (quota).


cooperlees:
🤖 **This review was written by Claude Code (Opus 5), posted by @cooperlees on its behalf.** Cooper asked for an AI review; findings below are mine, and were verified against the branch rather than read off the diff. Treat them as review input, not as his opinion.

Verified first: `cargo build`, `cargo test` (155 passed), `cargo clippy --all-features` (clean). The test-evidence claims in the description hold.

The recording points themselves look right. I traced every enabled collector through every branch in `lib.rs`, `machines.rs`, `boot.rs` and `verify.rs`: each one reaches exactly one assignment on every path, the write guards are temporaries dropped before the following `.await`s (no deadlock against the collectors that take the same lock), and both host and per-machine flattening are wired in correctly. The `Option` + `skip_serializing_if` design is the right call for making `count()` mean "enabled set". Three things are worth fixing before merge.

---

### 1. Boot-blame cache hits drop the gauge, which breaks the documented invariant

`update_boot_blame_stats` returns early on both cache paths — `src/boot.rs:236` (in-memory) and `src/boot.rs:250` (disk) — before any transport is recorded. The gauge is then **absent**, not 0 or 1.

That matters more than the edge-case note in the description suggests, because it is the default configuration doing the common thing:

- `monitord.conf:100` ships `cache_enabled = true`, and `config.rs:254` defaults it to `true`.
- One-shot under a systemd timer is the documented primary run mode.

So on a stock host, the first run after a boot writes the cache and emits `varlink_usage.boot_blame`; **every subsequent run in that boot emits nothing**. `count(varlink_usage)` changes between consecutive scrapes of the same host, and since boot blame is a collector that would report `1`, both numerator and denominator shrink — the adoption ratio is biased downward on exactly the hosts that have been up longest.

This contradicts `README.md:478` ("the present gauges are exactly the enabled set — no separate enabled-collectors counter is needed") without qualification.

CI can't catch it: `build_ci_configs` forces `cache_enabled = false`, so the integration test only ever exercises the collecting path.

Suggest recording the transport that produced the cached entry (persist it alongside the cached stats, replay it on a hit) so the gauge set stays stable across runs. If that's more than it's worth, the README claim needs the exception stated in it, not only in the PR body.

### 2. The enum serializes as a string in `json` / `json-pretty` and an integer in `json-flat`

`#[serde(rename_all = "lowercase")]` at `src/lib.rs:83` means the same metric has two different types depending on output format. Confirmed by running it:

```
json / json-pretty:  {"version":"varlink","units":"dbus"}
json-flat:           {"varlink_usage.version": 1, "varlink_usage.units": 0}
```

Every other stat enum that reaches output uses `Serialize_repr` so the value is the same integer in all three formats — `SystemdSystemState` (`src/system.rs:37`), `SystemdUnitActiveState` / `SystemdUnitLoadState` (`src/unit_constants.rs:17,45`). AGENTS.md calls this out as the convention ("`serde_repr` for integer mapping").

Anything parsing `json-pretty` against the documented 0/1 gauge gets a type mismatch. `Serialize_repr`/`Deserialize_repr` with `#[repr(u8)] Dbus = 0, Varlink = 1` would make `as_u64()` fall out of the repr too.

### 3. The container `units` gauge and the host `units` gauge don't mean the same thing

`src/machines.rs:369` records `Varlink` at the end of a block that unconditionally does per-unit D-Bus work on the way there: `collect_all_timers_dbus` (`machines.rs:326`) and `apply_oneshot_dbus_override` (`machines.rs:356`), both over the container's D-Bus connection. The host varlink path in `lib.rs` calls neither.

So a container reports `machines.<name>.varlink_usage.units = 1` while doing strictly more D-Bus work than a host that reports `0` because `apply_unit_details` failed (`src/lib.rs:469`). Both key shapes match a fleet-wide `varlink_usage` aggregation, so adoption is overstated on any host running containers.

I don't think the fix is to report `0` — the container genuinely used varlink for the bulk of the work. But the two gauges need either the same meaning or different names, and `VarlinkUsage::units`' doc comment should say which D-Bus calls still happen underneath a `1`.

---

### Smaller things

**The headline PromQL query can't match the emitted metrics.** `README.md` (new section), the `flatten_varlink_usage` doc comment in `src/json.rs`, and the PR description all give:

```promql
sum(varlink_usage) / count(varlink_usage)
```

But flat keys become individual Prometheus metric names (`README.md:315` says so explicitly) — `varlink_usage_units`, `varlink_usage_version`, and so on. There is no series named `varlink_usage`, so this selects nothing. It needs a name matcher:

```promql
sum({__name__=~"varlink_usage_.*"}) / count({__name__=~"varlink_usage_.*"})
```

Worth getting right in all three places, since it's the stated purpose of the change. (This is also where finding #3 bites — that matcher catches the per-machine keys too.)

**README sample isn't in sorted key order.** This PR adds the AGENTS.md rule "in sorted key order to match the `BTreeMap` flat output", then lists `varlink_usage.version` before `varlink_usage.verify`. `verify` sorts first (`i` < `s` at the 4th character).

**Host `MachineStats` is never reset between daemon iterations.** It's constructed once outside the loop at `src/lib.rs:291`, whereas per-container stats get a fresh `MachineStats::default()` each run (`src/machines.rs:154`). Today every enabled host collector rewrites its gauge each run so only boot blame goes stale — which is the intended behaviour per the description. But a future collector added with an early-return path would silently inherit a stale gauge instead of dropping it. Worth a note at the field.

**Gauge-write placement differs between the two files.** `lib.rs:492` sets `Dbus` *before* the D-Bus call; `machines.rs:386,404` set it *after*. Both work today. The `lib.rs` ordering is the more robust one — it survives the call returning `Err` — so it's the better one to standardise on before someone copies the wrong pattern into a new collector.

**Description nit:** it says the new assertion "replaces the weaker log-scrape fallback check", but `assert_no_varlink_fallback(varlink_log)` is still called at `varlink_integration_test.py:587`. Keeping both is right; just the description is off.

---

### Worth knowing, not a defect

`[machines] enabled = true` is the shipped default and the host `machines` gauge is hard-coded to `Dbus`, so a default deployment is permanently capped at 6/7 = 86% adoption until #37 lands. Documented, flagged only so it isn't later mistaken for a regression.

The test coverage here is genuinely good — the Rust test pinning the None-skips-the-gauge invariant plus five Python tests covering each failure mode of the new assertion is more than this kind of change usually gets. And the measured ~15ms networkd varlink cost note on `get_networkd_state` is the sort of thing that saves the next person an afternoon.

PR #223, opened 2026-09-19
Read all cgroup-derived service stats from cgroupfs (#221)

### LINKED ISSUE #221: Read all cgroup-derived service stats from cgroupfs instead of D-Bus/varlink
Seven `ServiceStats` fields describe a service's cgroup, and every one of them is a plain cgroup v2 file the kernel already exposes:

| `ServiceStats` field | cgroupfs source |
|---|---|
| `cpuusage_nsec` | `cpu.stat` → `usage_usec` |
| `memory_current` | `memory.current` |
| `memory_available` | `memory.max` / `memory.high` − `memory.current` |
| `tasks_current` | `pids.current` |
| `processes` | `cgroup.procs` (already done — see below) |
| `ioread_bytes` | `io.stat` → `rbytes` |
| `ioread_operations` | `io.stat` → `rios` |

Today these come from D-Bus unit properties (`src/units.rs:320-329` — `sp.memory_current()`, `sp.ioread_bytes()`, …).

## Why this blocks #37

There is **no varlink equivalent that works in practice.** `io.systemd.CGroup` is not an interface at all — `/run/systemd/report/io.systemd.CGroup` serves `io.systemd.Metrics`, and that socket emits only for paths where `cg_path_get_unit()` succeeds and stops at delegation boundaries (`cg_is_delegated()`), so it returns `NoSuchMetric` in our test container. `Unit.List`'s `runtime.CGroup` does carry these values, but only when the matching accounting flag is on.

So these seven fields are stuck on D-Bus, which is a hard blocker for the zero-D-Bus claim tracked in #37.

## systemd itself reads these files directly

`src/report/report-cgroup.c` in systemd parses `cpu.stat`, `io.stat`, `memory.current`, `memory.max/high/peak` and `pids.current` via `cg_get_keyed_attribute()` / `cg_get_attribute_as_uint64()`. There is no accounting API underneath to depend on — the accounting flags only control whether the *controller* is enabled, which is itself visible in `cgroup.subtree_control`.

Verified on Fedora Rawhide (systemd 262~rc3), reading a unit's cgroup with no systemd involvement at all:

```
cpu.stat        usage_usec 210697, user_usec 151702, system_usec 58995
memory.current  3829760
memory.peak     6471680
pids.current    1
```

## We already do this for one field

`count_cgroup_processes()` (`src/varlink_unit.rs:240`) walks `{fs_root}/sys/fs/cgroup{cgroup_path}` recursively, sums `cgroup.procs`, and folds in the main PID the way systemd does. The proposal is to generalise that into a small cgroupfs reader covering the rest of the table, used by both the varlink and D-Bus paths.

## Why it is worth doing

- Removes seven fields' dependency on D-Bus — otherwise a hard blocker for the zero-D-Bus goal in #37.
- `count_cgroup_processes()` already takes `fs_root` to prefix the mount for containers, so this works for machines **without** hitting the PID-namespace wall in #211: container unit cgroups are readable through `/proc/<leader>/root/sys/fs/cgroup/...`. That would make it the first per-machine stat family needing neither D-Bus nor varlink.
- File reads instead of IPC: fewer deps, smaller binary, better for the embedded goal.
- No version floor. cgroup v2 predates every varlink endpoint we depend on.
- Drops the dependency on per-unit accounting flags being enabled.

## Open questions for implementation

- The cgroup path itself still comes from systemd (`runtime.CGroup.Path` on varlink, the `ControlGroup` property on D-Bus). Cheap either way, and already present in the varlink reply.
- `memory_available` is derived rather than read — confirm the current D-Bus `MemoryAvailable` semantics before reimplementing so output does not silently change.
- `io.stat` only exists when the `io` controller is enabled on the subtree. Confirmed on Rawhide: with `IOAccounting=yes`, systemd adds `io` to `system.slice/cgroup.subtree_control` and `io.stat` appears; without it, the file is absent. Decide whether absent means 0 or a missing key, matching today's behaviour.
- cgroup v1 hosts: support, or read v2 only and fall back to D-Bus?
- The parity test currently excludes `cpuusage_nsec`, `memory_current` and `memory_available` as volatile (`varlink_integration_test.py:80-90`). Moving both paths to a single source may let some of those exclusions go away.

Refs #37, #211

cooperlees:
Fixes https://github.com/cooperlees/monitord/issues/221 — all seven cgroup-derived `ServiceStats` fields now come from cgroupfs instead of D-Bus/varlink. Helps https://github.com/cooperlees/monitord/issues/37 (zero-D-Bus goal): these fields had no working varlink equivalent, so they were a hard blocker there.

New `src/cgroup.rs`: async (tokio::fs + join!, zero new deps) cgroup v2 reader for all seven cgroup-derived `ServiceStats` fields. Semantics taken from systemd source (`cgroup.c`, `dbus-unit.c`, `basic/procfs-util.c`) and verified live byte-exact (details in the commit message). Both collectors share the one reader, joined in parallel with their IPC, with per-field fallbacks so cgroup v1 hosts see no new failure mode. Per-unit accounting flags intentionally no longer consulted.

Verified: 170 unit tests pass, fmt/clippy clean, no perf regression (60.1→59.7ms on 15 services), Rawhide container integration test passes with a new all-accounting-on fixture service reporting byte-identical real values on both paths (92 keys identical).


cooperlees:
> Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.

@copilot When do I get more review credit - How can I check?

cooperlees:
> Posted by Claude Code on behalf of @cooperlees — @cooperlees asked me to review this PR and leave the findings here. Treat this as a review comment from an assistant, not a human sign-off.

## Summary

The core idea is sound and the systemd semantics are right. I checked every claim in the PR/commit body against systemd `main` sources and against a live systemd 259 box (Ubuntu 24.04, cgroup v2), and the numbers hold up.

**Verified against systemd source** (fetched `main`, not from memory):
- `unit_get_cpu_usage_raw()` in `src/core/cgroup.c` → `cg_get_keyed_attribute_uint64(..., "cpu.stat", "usage_usec")` × `NSEC_PER_USEC`. ✅ matches `parse_cpu_stat`.
- `unit_get_memory_available()` → the `do/while (u = UNIT_GET_SLICE(u))` walk, `MIN3(unit_limit, memory_max, memory_high)`, `LESS_BY(unit_limit, current)`, `continue` on an unlimited level *after* refreshing `current`, `break` on `available == 0`, and `unit_limit = physical_memory()` for `SPECIAL_ROOT_SLICE`. ✅ `fold_memory_available` mirrors all of it, including the "previous current propagates as lower bound" comment.
- `procfs_memory_get()` in `src/basic/procfs-util.c` → `*ret_used = mem_total - mem_available`. ✅ matches `read_host_memory`.
- `unit_get_io_accounting_raw()` → skips the `major:minor` head, `startswith(word, "rbytes=")`, sums across devices. ✅ matches `parse_io_stat`.
- `unit_get_tasks_current()` → `pids.current`. ✅
- One mismatch, see finding 6 below.

**Verified live** (`alloy.service`, systemd 259, all values from one `monitord` run vs `busctl` moments later):

| field | monitord (cgroupfs) | D-Bus |
|---|---|---|
| `cpuusage_nsec` | 6173997857000 | 6173997857000 (byte-exact) |
| `memory_current` | 144510976 | 144510976 (byte-exact) |
| `tasks_current` | 21 | 21 |
| `memory_available` | 4570402816 | 4568592384 (≈, `MemAvailable` jitter between reads) |
| `ioread_bytes` | 723297280 | `[not set]` — the documented `IOAccounting=no` improvement |
| `processes` | 1 | `cgroup.procs` = `1088` = MainPID ✅ |

`cargo fmt --check` clean, `cargo test` 170/170 pass, `cargo clippy` adds no new warnings (15 pre-existing).

**On "are we reinventing an existing async cgroupfs crate?"** — I checked crates.io. No, the hand-roll is justified:

| crate | downloads | async? | extra deps pulled in |
|---|---|---|---|
| `cgroups-rs` 0.5.1 (kata) | 6.3M | ❌ sync | `bit-vec`, `libc`, `log`, `nix`, `thiserror`, `zbus` |
| `cgroupfs` 0.9.0 (Meta/below) | 260k | ❌ sync (`openat` fd-based) | `below-common`, `nix`, `openat`, `serde`, `thiserror` |
| `libcgroups` 0.7.0 (youki) | 196k | ❌ sync | `oci-spec`, `pathrs`, `procfs`, `fixedbitset`, … |
| `cgroups-fs`, `controlgroup` | tiny / unmaintained (2020) | ❌ sync | — |

There is no async cgroup v2 reader on crates.io. All three serious candidates are `std::fs`/`openat` and would need `spawn_blocking`, and none of them models `memory_available` (that's a systemd concept, not a kernel file) — which is where most of the complexity in `cgroup.rs` actually lives. So the module doc's justification is accurate; the only thing I'd tighten is that it names "`cgroups-rs` and friends" without saying which, and `cgroupfs` (Meta) is the closest miss and worth naming since it's the one a reviewer will reach for.

---

## Findings

### 1. `memory_available` never falls back — the "cgroup v1 → no new failure mode" guarantee doesn't hold for this one field 🔴

`src/cgroup.rs:372-397` + `fold_memory_available` at `:270-293`.

The root level always binds, so `read_memory_available` returns `Some(host MemAvailable)` even when the unit's cgroup directory doesn't exist at all. Every other field goes `None` → D-Bus/varlink fallback; this one silently doesn't. The PR's own test encodes the behaviour:

```rust
// src/cgroup.rs:612 — test_read_service_cgroup_missing_dir
// …except memory_available, whose root level always binds via
// the host /proc/meminfo.
```

Consequences:
- **cgroup v1 host**: no v2 files anywhere, so a unit with `MemoryMax=512M` reports host `MemAvailable` instead of falling back to the correct `ServiceProxy::memory_available()`. That contradicts both the module doc (`:64-72`) and `README.md:820`.
- **Race**: a unit that stops between the `ControlGroup` read and the fs read gets host `MemAvailable` while its five siblings fall back — an internally inconsistent row.

Suggested fix — only let the root level bind when the unit's own level was actually readable:

```rust
async fn read_memory_available(fs_root: &str, cgroup_path: &str) -> Option<u64> {
    // …existing walk…
    // Nothing readable at the unit's own level means no cgroup v2 here:
    // report no data so the caller's IPC fallback takes over, like every
    // other field.
    levels.first()?.current?;
    let root = read_host_memory()?;
    fold_memory_available(&levels, root)
}
```

and update `test_read_service_cgroup_missing_dir` to assert `memory_available == None`.

### 2. The D-Bus fallback is all-or-nothing, and the comment says the opposite 🟡

`src/units.rs:352-387`.

```rust
// Only missing fields pay for a fallback call; the hot path (cgroup
// present, the common case for running services) issues none.
```

The `match` has exactly two arms: all-six-`Some`, or *everything else* → `tokio::join!` of all six `ServiceProxy` getters. One absent `io.stat` (io controller not enabled on the subtree) costs six property round-trips per service per cycle, four of which are thrown away. Also worth noting the D-Bus path now costs **2–3 sequential round-trip phases** per service (`ControlGroup`+`MainPID`, then the 9-prop batch, then the fallback batch) where it used to be a single 16-prop `join!` — the fallback phase is the avoidable one.

Per-field, one round, same shape:

```rust
let (cpuusage_nsec, ioread_bytes, ioread_operations, memory_current, memory_available, tasks_current) = tokio::join!(
    async { match cgroup_stats.cpu_usage_nsec { Some(v) => Ok(v), None => sp.cpuusage_nsec().await } },
    async { match cgroup_stats.io_read_bytes  { Some(v) => Ok(v), None => sp.ioread_bytes().await } },
    // …
);
```

As a bonus the `?` then only applies to properties you actually needed, instead of failing the whole service because a property you already had from cgroupfs errored.

### 3. `read_host_memory()` blocks the runtime — the exact thing the module doc says it avoids 🟡

`src/cgroup.rs:308-332` uses `std::fs::read_to_string("/proc/meminfo")` inside an async call path, while `:44-52` argues against the ecosystem crates precisely because "blocking the tokio runtime for every service on every cycle is exactly what this collector avoids." Two fixes, both worth doing:

- `tokio::fs::read_to_string` (it's already imported in this module).
- It's host-global and identical for every service in a cycle, yet it's re-read once per service (and once per ancestor walk). Read it once in `parse_unit_state` / `apply_unit_details` and pass `(total, used)` into `read_service_cgroup`. At 15 services that's 15 procfs reads → 1.

### 4. Dead code: `count_cgroup_processes` and `cgroup::UNSET` 🟡

- `src/varlink_unit.rs:246` — the commit message says "compat wrapper kept", but `grep` finds zero callers in-tree; `varlink_units.rs` now calls `cgroup::read_service_cgroup` directly. It's `pub`, so no `dead_code` warning surfaces it. Worse, it now does the full seven-field read — `cpu.stat`, `memory.current`, `io.stat`, `pids.current`, the whole ancestor walk *and* `/proc/meminfo` — to return `.processes`. Please delete it, or if it's kept as public API, give it its own cheap path.
- `src/cgroup.rs:113` — `pub const UNSET` is never referenced; `varlink_unit.rs:33` has its own private `const UNSET`. Pick one (exporting from `cgroup` and deleting the private copy seems right) rather than shipping both.

### 5. Three mangled doc comments 🟡

These look like move/merge artifacts:

- `src/varlink_unit.rs:233-245` — the old doc block was left in place above the new one, so one comment now has two `/// Count the processes in a unit's cgroup, including nested ones.` headers with the stale D-Bus paragraph sandwiched between.
- `src/cgroup.rs:334-350` — the paragraph describing `read_memory_available` ("Compute `memory_available` for one service: … walk the ancestor slices …") is attached to **`read_memory_level`**, and then a second `/// One level's …` header follows. Net effect: `read_memory_level` has someone else's docs and `read_memory_available` (`:372`) has none.
- `src/cgroup.rs:211-212` — `keyed_u64`'s doc says "`io.stat` uses `rbytes=123` — hence the two prefixes", but the fn only ever builds one `"{key} "` prefix and is never called for `io.stat` (`parse_io_stat` does its own `split_once('=')`).

### 6. `processes` drops the control PID, where `GetProcesses` includes it 🟡

`src/core/dbus-unit.c`, `bus_unit_method_get_processes()`:

```c
/* The main and control pids might live outside of the cgroup, hence fetch them separately */
PidRef *pid = unit_main_pid(u);
if (pidref_is_set(pid)) { r = append_process(reply, NULL, pid, pids); … }

pid = unit_control_pid(u);
if (pidref_is_set(pid)) { r = append_process(reply, NULL, pid, pids); … }
```

`count_processes` folds in only `main_pid`. So during `ExecStartPre`/`ExecReload`/`ExecStop`, a control process living outside the cgroup is now uncounted where the old `get_processes()` counted it — a 1-off that the module doc's "folded in the same way systemd does" claim doesn't cover. `ServiceProxy::control_pid()` is one more entry in the existing `join!`, and varlink has `runtime.service.control_pid`. (Both go into the same `HashSet`, so dedup is free.)

### 7. Nits 🟢

- **Duplicate + serialised reads** (`src/cgroup.rs:134-157`): the unit's own `memory.current` is read twice — once in the `join!` batch, again inside `read_memory_level` for level 0. And `read_memory_available` is `.await`ed *after* the `join!` rather than as a sixth arm of it, with the ancestor walk itself serial per level (`levels.push(read_memory_level(&dir).await)`). All cheap, but folding the walk into the same `join!` and reusing the already-read `memory.current` is close to free.
- **`read_host_memory` bails on the first odd line** (`:313-315`): `line.split_once(':')?` and `.parse().ok()?` abandon the whole file, where systemd's `procfs_memory_get()` does `else continue;` on lines it doesn't recognise. In practice `MemTotal`/`MemAvailable` are lines 1 and 3 so the early `break` saves you, but `continue` is both more faithful and free.
- **Varlink path has no empty-cgroup-path guard**: `units.rs:313` correctly guards with `(!control_group.is_empty())`, but `varlink_units.rs:739` passes `path` straight through. An empty `runtime.CGroup.Path` would make `dir` = `{fs_root}/sys/fs/cgroup` and send `count_processes` walking the **entire host cgroup tree**, counting every PID on the box, plus reading the root `cpu.stat`/`io.stat`. Cheapest fix is one guard inside `read_service_cgroup` itself (`if cgroup_path.is_empty() || cgroup_path == "/" { return CgroupStats::default(); }`), which covers both callers and the public API.
- **Possible integration-test flake** (`varlink_integration_test.py`, `assert_cgroup_fixture_values`): `ioread_bytes`/`ioread_operations` are asserted `== 0` exactly. `sleep infinity` does no IO of its own, but the `execve` can charge page-cache misses to the fixture's cgroup on a cold container — the io counters aren't as deterministic as `processes`/`tasks_current`. Consider asserting "not `NOT_SET`" for the two io fields and keeping the exact check for the two that really are deterministic.
- **`fold_memory_available` vs systemd on an unreadable `memory.current`** (`:270-293`): systemd initialises `current = 0`, so a first level with a limit but an unreadable current yields `LESS_BY(limit, 0) == limit`; this skips the level entirely, i.e. reports *more* available. Truly an edge case, but the doc claims an exact mirror.
- **README**: `README.md:820`'s "falls back to its D-Bus property when cgroupfs has no data (e.g. cgroup v1 hosts)" needs the `memory_available` caveat if finding 1 isn't fixed.

---

Nothing here is a blocker on the approach — 1 is the one I'd want fixed before merge, 2–6 are small and mechanical, 7 is optional.

cooperlees:
All seven findings addressed in 52853b3, just pushed. Thanks for the thorough review — every claim re-checked against systemd source. Disposition:

1. **`memory_available` never fell back** 🔴 → fixed as suggested. Root level only binds when the unit's own level was readable; missing cgroup now reports `None` → IPC fallback like every other field. Missing-dir test asserts all-`None`.
2. **All-or-nothing D-Bus fallback** 🟡 → fixed. Per-field fallback in the same single `join!` batch; a failing property fails only its own field. D-Bus path is back to two round phases instead of three.
3. **`read_host_memory` blocked the runtime** 🟡 → fixed. Async `tokio::fs`, read once per cycle and shared (`HostMemory` param on both paths).
4. **Dead code** 🟡 → fixed. `count_cgroup_processes` wrapper deleted; single `UNSET` definition kept.
5. **Mangled docs** 🟡 → fixed. Headers re-attached, walk renamed `read_ancestor_levels`, `keyed_u64` doc corrected, crate survey (cgroups-rs / Meta cgroupfs / libcgroups) folded into the justification.
6. **Dropped control PID** 🟡 → fixed. `count_processes` takes main + control PIDs; D-Bus path fetches `control_pid()`, varlink `ServiceRuntime` gains `ControlPID`. Verified live on `alloy.service`.
7. **Nits** 🟢 → all taken: single `memory.current` read, `continue`-style meminfo parse, empty-path guard inside `read_service_cgroup`, IO assertions check present-not-`MAX`, unreadable-current contributes `limit − 0` (with test), README updated.

Verification after the fixes: `cargo test` 174/174, fmt clean, clippy 16 warnings (same 16 pre-existing on main), no perf regression (56.1ms vs 60.1ms baseline on 15 services), Rawhide integration test passes with byte-identical fixture values on both paths, 92 keys identical.

cooperlees:
> Posted by Claude Code on behalf of @cooperlees.

Re-checked `52853b3` — all seven findings are addressed and verified: `cargo fmt` clean, 174/174 tests (+4), clippy unchanged at 16 with zero in `src/cgroup.rs`, and a live re-run against `busctl` on systemd 259 matches (`memory_current` byte-exact, `tasks_current` 21, `cpuusage_nsec` within 82µs, `memory_available` within 286KB of D-Bus — `MemAvailable` jitter; MainPID 1088 / ControlPID 0 with `processes=1`, so the zero is correctly filtered).

One item only partly landed: **finding 3's "read once per collection cycle" half, on the D-Bus path.**

`src/units.rs:312-317` calls `read_host_memory()` inside `parse_service`, so it runs once per *service*, not once per cycle — while the comment immediately above says otherwise:

```rust
// `host_memory` is read once per
// collection cycle (see `parse_unit_state`) and shared by every unit.
```

`parse_unit_state` has no such read; `grep read_host_memory src/` finds only `units.rs:316`, `varlink_units.rs:726` and the definition. The varlink path does it correctly (once, before the `join_all` at `varlink_units.rs:726`), so the two paths are asymmetric.

The blocking half *is* fixed — it's `tokio::fs` now, so this is minor in practice — but the comment actively misdirects. The hoist is small: read it once in `parse_unit_state` alongside the `semaphore` / `parent_span` setup around `src/units.rs:836-844`, then thread `Option<HostMemory>` into `parse_service` the same way `fs_root` already is:

```rust
// once, outside the per-unit JoinSet
let host_memory = crate::cgroup::read_host_memory().await;
// …
for unit in listed_units {
    let fs_root = fs_root.to_string();
    // host_memory is Copy ((u64, u64)), so no clone needed
    join_set.spawn(async move {
        // …
        parse_service(&connection, &unit.name, &unit.unit_object_path, &fs_root, host_memory).await
    });
}
```

At 15 services that's 15 procfs reads → 1, and it makes the comment true. Either that or drop the "(see `parse_unit_state`)" claim from the comment.

cooperlees:
Good catch — fixed in b96cb5a, just pushed. `parse_unit_state` now reads `host_memory` once alongside the semaphore/`parent_span` setup and threads `Option<HostMemory>` into `parse_service` the same way `fs_root` already is, so both paths are symmetric (15 procfs reads → 1 at 15 services) and the comment is true. 174/174 tests, fmt clean, clippy unchanged at 16, local values re-verified after the move.

PR #224, opened 2026-09-19
Connect to D-Bus lazily instead of eagerly at startup

cooperlees:
Host item 1 on the road to zero-D-Bus in https://github.com/cooperlees/monitord/issues/37: `stat_collector` built the system bus connection up front and errored out on failure, so monitord could not start without a bus regardless of which collectors were enabled. This blocked the bus-blocked proof run.

Shared `Arc<tokio::sync::OnceCell<zbus::Connection>>`, seeded from `maybe_connection` for library callers, connected on first use. Each collector tries varlink/fs/procfs first and only resolves the cell on a D-Bus code path (fallbacks included) — a missing bus is that collector's ordinary per-collector error, logged never fatal. Daemon reuse preserved; a failed cycle returns `None` so the next cycle reconnects.

Verified: dead-bus run (`unix:path=/nonexistent/bus_socket`) with varlink version+system-state+procfs pid1 exits 0 with both gauges at 1 (old code hard-errored); a D-Bus-needing run against the same dead bus emits JSON with per-collector `success=0` instead of crashing. `cargo test` 174/174, fmt clean, clippy output identical to main.


cooperlees:
@ikruglov + @aleivag so close now to varlink only :)

cooperlees:
> Posted by Claude Code on behalf of @cooperlees — @cooperlees asked me to review this PR and leave the findings here. This is an assistant review, not a human sign-off.

## Summary

The shape is right and the headline win is real. I built the branch and confirmed the core claim live: with `dbus_address = unix:path=/nonexistent/bus_socket` and eight D-Bus collectors enabled, monitord **exits 0** and emits JSON with per-collector failures instead of hard-erroring at startup:

```
Collector 'boot_blame' failure ... Collector 'version' failure   (8 collectors, each once)
exit=0
```

`cargo fmt` clean, 174/174 tests pass, clippy 16 warnings — unchanged.

I also re-checked @copilot-pull-request-reviewer's three findings independently rather than taking them at face value. **All three are valid**, and I have a live repro for the first. Details below, plus three it didn't catch.

---

## Confirming Copilot's findings

### 1. Eager `dbus_connection` in `verify` and `boot_blame` defeats the PR's own goal 🔴

Copilot flagged this at `src/lib.rs:605` and `:615`. Confirmed, and it's the one I'd block on.

The PR body says "Each collector tries varlink/fs/procfs first and only resolves the cell on a D-Bus code path (fallbacks included)." That holds for `version`, `networkd`, `system_state` and `units` — I checked each: they attempt varlink, `return Ok(())` on success, and only reach `dbus_connection` on the fallback path. ✅

It does **not** hold for the two collectors that make the varlink decision *inside* the collector function:

- **`verify`** (`src/lib.rs:615`) — `update_verify_stats` branches on `varlink_enabled` at `src/verify.rs:221` and only touches `connection` in the fallback (`:240`) or non-varlink (`:248`) arm. But `lib.rs` resolves the cell before calling it.
- **`boot_blame`** (`src/lib.rs:604`) — `update_boot_blame_stats` returns `Ok(())` early on an in-memory cache hit (`src/boot.rs:283-286`) and on a disk cache hit (`:292-306`), neither of which touches `connection`, and it has its own varlink path at `src/boot.rs:357-360`.

**Live repro for `verify`**, same config twice, only `dbus_address` differs:

```
A) live bus:
W src/verify.rs:234  Varlink verify enumeration failed, falling back to D-Bus: I/O error: No such file or directory
D src/lib.rs:674     collector 'verify' ... elapsed=233.5ms

B) dead bus:
E src/lib.rs:645     Collector 'verify' failure: D-Bus connection error: Failed to connect to address `unix:path=/nonexistent/bus_socket`
D src/lib.rs:674     collector 'verify' ... elapsed=0.5ms (FAILED)
```

The `verify.rs:234` warning is **absent** in run B — varlink was never attempted. It died at `dbus_connection` in 0.5ms. On a systemd ≥261 host with `/run/systemd/report/io.systemd.Manager` present, verify would have succeeded over varlink on a bus-less box; this PR makes it fail. (My box is systemd 259, so `/run/systemd/report/` doesn't exist and varlink falls back either way — which is exactly why the missing warning is the decisive signal.)

Fix: make `dbus_connection` and `DbusCell` `pub(crate)` and pass the cell + timeout into these two, resolving at the two call sites that actually need a `&Connection` — `verify.rs:240`/`:248` and inside `boot.rs::collect_and_cache`. That also keeps the "only on a D-Bus code path" invariant enforced in one place rather than re-argued per collector.

### 2. Daemon mode never resets the cell, and the new comment says it does 🟡

Copilot flagged `src/lib.rs:724`/`:727`. Confirmed:

```rust
if !config.monitord.daemon {
    break;                         // :709 — the ONLY exit from the loop
}
// …sleep, next cycle…
}
let conn = if had_error { None } else { dbus_cell.get().cloned() };   // :724-728
```

The reset is after the loop, and in daemon mode the loop never exits — so a failed cycle never drops the connection and the next cycle reuses the same (possibly dead) bus. `src/main.rs:41` discards the return value entirely, so today the `None` only ever reaches a library caller doing repeated one-shot runs.

To be fair to the PR: this is **not a regression** — pre-PR, `sdc` was likewise built once outside the loop and reused forever. What's new is the comment and the PR body ("a failed cycle returns `None` so the next cycle reconnects") claiming a self-healing property that daemon mode doesn't have, and daemon mode is precisely where a bus restart matters. Either fix the comment, or take Copilot's suggestion and make it true — now cheap, since it's `let mut dbus_cell` plus a reset before the sleep:

```rust
if had_error {
    // Drop the shared connection so the next cycle reconnects rather than
    // reusing a bus that may have gone away mid-run.
    dbus_cell = Arc::new(tokio::sync::OnceCell::new());
}
```

(Each iteration `Arc::clone`s into its tasks at spawn time, so reassigning the local is enough — the next cycle's spawns pick up the new cell.)

### 3. README overstates the no-bus guarantee for networkd 🟢

Copilot's nit at `README.md:761`. Confirmed: `lib.rs` passes `None` for `maybe_network_int_to_name`, so `parse_interface_state_files` always takes the `None` arm at `src/networkd.rs:365` and calls `get_interface_links(connection)` → `Manager.ListLinks`. An enabled networkd collector without varlink reads state *files* but still needs the bus for the ifindex→name map. Worth one clause in the README so "fs paths never connect" isn't read as covering networkd.

---

## Not caught by Copilot

### 4. The `DbusCell` doc comment is wrong about `tokio::sync::OnceCell` 🟡

`src/lib.rs:289-291`:

> Concurrent first-users race `get_or_try_init`, but `zbus::Connection` is cheap to clone, so losers drop their spare and everyone shares one connection.

Tokio's `OnceCell` doesn't work that way — it serializes initializers behind a semaphore. Per the docs: *"If some other task is currently working on initializing the `OnceCell`, this call will wait for that other task to finish, then return the value that the other task produced."* There is no race and no spare connection to drop; that description fits `std::sync::OnceLock` / `once_cell::race`, not this type. The second half of the comment is correct — *"If the provided operation returns an error, is cancelled or panics, the initialization attempt is cancelled. If there are other tasks waiting for the value to be initialized, one of them will start another attempt"* — so a failed connect is genuinely not cached.

The consequence is worth stating because it bites the exact scenario this PR targets: **not-cached + serialized means a dead bus costs one connect attempt per D-Bus collector, one after another.** Measured on the branch — 8 collectors, 8 attempts:

```
$ monitord -c dead-bus.conf 2>&1 | grep -c "Failed to connect to address"
16          # 8 attempts × 2 lines of error chain each
```

For `ENOENT` that's free (the whole run finished in milliseconds). But `zbus::connection::Builder::build()` isn't bounded by `method_timeout` — that only covers method calls — so a socket that exists and hangs (wedged broker, the "bus-blocked proof run" case) would serialize N blocking connects per cycle, where the old code blocked exactly once at startup. At minimum the comment should say so; a connect-side timeout would be better.

### 5. The transport gauge is set to `Dbus` before the connect that then fails 🟢

`version` (`:397`), `networkd` (`:435`), `system_state` (`:481`), `units` (`:551`) and `machines` (`:568`) all write `varlink_usage.* = Some(CollectorTransport::Dbus)` and *then* resolve the cell. On a bus-less host that reports "Dbus" for collectors that produced nothing. Since #219 added these gauges for Grafana adoption reporting, this skews the metric on exactly the hosts this PR exists to enable — and it's newly reachable, because a failed connect is no longer fatal. Setting the gauge after the collector succeeds would keep it honest.

### 6. No test covers the headline behavior 🟡

174/174 pass, but that's the same 174 as main — nothing new is asserted, and the whole PR is a behavioral change. The verify bug in finding 1 shipped through a green run.

Two cheap additions:
- A unit test on `dbus_connection` directly: bogus address → `Err` twice in a row, with `cell.get().is_none()` in between. That pins the "failure is not cached" property the design leans on.
- A case in `varlink_integration_test.py` that runs varlink-only against a dead `dbus_address` and asserts exit 0 with every enabled collector at `success=1`. That is the actual claim of the PR, and it would have failed on `verify` today.

---

Finding 1 is the blocker for me; 2 and 4 are comments that promise more than the code delivers; 3, 5, 6 are small. Nice result overall — `exit=0` on a dead bus is a real step toward #37.

cooperlees:
All six findings addressed in f1cfc72, just pushed — 5 fixed, 1 declined with reasoning:

1. **Eager resolution in verify/boot_blame** 🔴 → fixed. `DbusCell` flows into both collectors; they resolve only on the D-Bus arms. Live proof on a dead bus: varlink verify now logs the varlink-attempt warning *before* the connect error, instead of dying in 0.5ms without trying.
2. **Daemon cell reset** 🟡 → fixed. Fresh cell on failed cycles before the sleep; post-loop comment corrected.
3. **README networkd exception** 🟢 → fixed, one clause.
4. **OnceCell doc** 🟡 → fixed. Rewritten honestly (serialized init, uncached errors, wedged-broker sequential-connect cost).
5. **Gauge set before failed connect** 🟢 → declined. Setting `Dbus` only after success would inherit the previous cycle's gauge on failure — the stale-inheritance hazard the `VarlinkUsage` docs explicitly warn about. `Dbus` + `success=0` honestly answers "did this use varlink: no".
6. **No headline test** 🟡 → fixed. `test_dbus_connection_failure_is_not_cached` unit test + a dead-bus integration case (exit 0, all `success=1`; networkd/verify/machines excluded with reasons).

Verification: 175/175 tests, fmt clean, clippy identical to main, full integration suite green including the new dead-bus case on systemd 262.

PR #225, opened 2026-09-20
Add [varlink] no_fallback mode: fail loudly instead of falling back

cooperlees:
Follow-up to #224 and the next verification step for https://github.com/cooperlees/monitord/issues/37: a shared enforcement point (`src/varlink_fallback.rs::report_varlink_failure`) that every varlink-first collector funnels failures through. Default behavior unchanged (warn + fallback); with `[varlink] no_fallback = true` the failure becomes the collector's error, so CI proves a collector set is varlink-clean rather than silently D-Bus-served.\n\nWired through version, networkd, system-state, units (metrics + details), verify, boot blame, and all four container fallbacks (naming the machine). The dead-bus integration case now sets `no_fallback=true`, doubling the proof — any attempted fallback would fail its bus connect.\n\nVerified: 177 unit tests, fmt clean, clippy identical to main. Local: passes with working varlink, fails loud on broken varlink. Rawhide container: full suite green including the dead-bus + no_fallback run.


cooperlees:
> Posted by Claude Code on behalf of @cooperlees — @cooperlees asked me to review this PR and leave the findings here. This is an assistant review, not a human sign-off.

## Summary

The design is right: one enforcement point every varlink-first collector funnels through beats scattering `if no_fallback` across seven call sites, and the default path is genuinely untouched. Everything below was checked by building and running the branch, not by reading alone.

**Verified good:**
- `cargo fmt` clean, 177/177 tests, clippy 16 warnings — unchanged.
- The wiring claim holds. I grepped for leftover `warn!(… falling back …)` sites across `src/`: every one now routes through `report_varlink_failure`, and the only remaining matches are comments.
- Config plumbing is correct — `read_config_optional_bool` means a missing key keeps the derived `Default` (`false`), and `KNOWN_SECTION_KEYS` is updated so the new key isn't warned as a typo.
- The feature does what it says: `Collector 'units' failure: varlink units failed and [varlink] no_fallback=true forbids the D-Bus fallback: I/O error: No such file or directory (os error 2)`.

Six findings, the first two substantive.

---

### 1. `no_fallback = true` is a silent no-op when varlink is off 🟡

`no_fallback` only fires from inside a `use_varlink(...)` branch, so turning varlink off anywhere disables the check rather than tripping it. With `[varlink] enabled = false` and `no_fallback = true`:

```
success:       {'system_state': 1, 'units': 1, 'version': 1}
varlink_usage: {'system_state': 0, 'units': 0, 'version': 0}     # 0 == D-Bus
monitord exit=0
warnings mentioning no_fallback: 0
```

Every collector served by D-Bus, everything green, not a word about it. The same holds for a per-section `[units] varlink = false`.

The flag proves *"no fallback was taken"*, which is not the same as *"varlink was used"* — and the PR body claims the second ("CI proves a collector set is varlink-clean rather than silently D-Bus-served"). Someone who sets `no_fallback = true` to get that guarantee, and has a section toggle off, gets a fully green all-D-Bus run instead.

Cheapest fix is a config-load warning:

```rust
if config.varlink.no_fallback && !config.varlink.enabled {
    warn!("[varlink] no_fallback=true has no effect while [varlink] enabled=false");
}
```

The thorough version asserts `varlink_usage.* == Varlink` for every enabled varlink-capable collector at end of run, which is the property CI actually wants.

### 2. A `no_fallback` failure still exits 0 🟡

Measured with a real varlink failure under `no_fallback = true`:

```
success gauges: {'system_state': 1, 'units': 0, 'version': 1}
monitord exit=0
```

"Fail loudly" is loud in the log line and the `success` gauge, but not in the process exit status. `assert_dead_bus_run` only catches it because it parses `collector_timings.*.success` out of the JSON — a CI gate written the obvious way (`monitord -c strict.conf && echo clean`) passes silently.

Given `no_fallback` is explicitly a verification mode, a nonzero exit when it trips seems right and wouldn't touch the default path, where per-collector errors are deliberately non-fatal. If that's too big a contract change for this PR, the README should say plainly that CI must assert on the `success` gauges.

### 3. `boot.rs` discards the error message the module went to the trouble of writing 🟡

`src/boot.rs` is the only call site using `.map_err(|e| anyhow!("{:?}", e))?` instead of plain `?`. `{:?}` selects the `#[derive(Debug)]` rendering, not the `#[error(...)]` one. Both, printed side by side:

```
via ?              : varlink boot blame failed and [varlink] no_fallback=true forbids the D-Bus fallback: boom
via map_err {:?}   : FallbackForbidden { collector: "boot blame", fallback: "D-Bus", source: boom }
```

So boot blame is the one collector whose no-fallback error doesn't read like the others. I confirmed plain `?` compiles there (`From<NoFallbackError> for anyhow::Error` applies, same as every other call site), so the `map_err` is redundant as well as lossy — just drop it.

### 4. Log casing regression: `falling back to d-bus` 🟢

`fallback.to_lowercase()` at `src/varlink_fallback.rs:57` lowercases the proper noun:

```
Varlink units failed, falling back to d-bus
```

It was `falling back to D-Bus` before this PR. It doesn't break `find_fallbacks()` in the integration test (that matches on `"falling back"`), but #222 was specifically about making these fallback lines read well. Either drop the `.to_lowercase()` and keep `"D-Bus"`/`"file-based"` as passed, or pass already-lowercased strings at the call sites — the transform in the middle is what makes it surprising.

### 5. README isn't updated for the new key 🟢

`monitord.conf` documents `no_fallback`, but `README.md` doesn't, and its "### Enabling Varlink" section currently tells the reader the opposite:

> When varlink is enabled, monitord will attempt to collect stats via the varlink APIs first, automatically falling back to D-Bus or file-based collection when a varlink socket is unavailable (e.g., older systemd versions).

That paragraph is where someone would look for this. The per-collector "Falls back to D-Bus if the socket is unavailable" bullets further down in "Metrics collected via Varlink" are now conditional too.

### 6. Worth stating what `no_fallback` can't see 🟢

Two gaps that are fine as behavior but shouldn't be papered over by the "varlink-clean" phrasing:

- **D-Bus-only collectors are invisible to it.** `dbus_stats` and the `machines` enumeration have no varlink path, so they aren't fallbacks and never report one — they just use the bus. (The dead-bus integration run sidesteps this by disabling them, and its comment says so, which is good.)
- **Partial varlink data isn't a fallback.** The metric-parse warnings in `varlink_units.rs` (`has non-string value`, `has unrecognized value`, `has non-integer value`) skip the metric and carry on rather than falling back, so a `no_fallback` run can be "clean" while quietly under-reporting. The integration test's other assertions cover completeness; the claim just shouldn't rest on `no_fallback` alone.

---

1 and 2 are the ones I'd want addressed — together they mean the guarantee is weaker than the PR body states, in two independent ways, and both are cheap to close. 3–6 are small.

cooperlees:
All findings addressed in d681d03 — 6 fixed, 3 declined with reasoning:

**Fixed**
- **Duplicate `no_fallback` key (Copilot moderate).** Confirmed live and serious: the container file carried both `true` and `false`, and I proved with a scratch binary that configparser last-wins — so the "dead-bus + no_fallback" proof ran with the flag off. Now replaces the existing block; the regenerated container file carries exactly one assignment.
- **`{:?}` on `NoFallbackError` (Copilot low, Claude 3).** Plain `?` now, matching every other call site.
- **Silent no-op when varlink off (Claude 1, cheap half).** Config load warns. Verified live.
- **Casing regression (Claude 4).** `.to_lowercase()` dropped; names pass in display form.
- **README missing the key (Claude 5).** Documented in Enabling Varlink, including the exit-status contract and scope limits below.
- **Unstated blind spots (Claude 6).** Module + README now say D-Bus-only collectors and warn-and-continue parses are outside the flag's view.

**Declined**
- **Nonzero exit on tripped `no_fallback` (Claude 2).** `main` returns `()` unconditionally and daemon mode must survive collector failures — that's the documented philosophy, and a wedged socket at 3am shouldn't kill the daemon. Loudness comes via log + `success` gauges, which is what CI asserts; README now states this plainly.
- **End-of-run gauge assertion (Claude 1, thorough half).** Needs a per-collector expected-transport model that would rot, and section-off + `no_fallback`-on is a legitimate migrate-one-at-a-time workflow — erroring there breaks what the toggles were built for.
- **Reject D-Bus-stamped cache entries (Copilot boot-cache nit).** A replayed entry's gauge honestly says `Dbus`; re-collecting defeats the cache. The claim is about collection, and the gauge answers it.

Verification: 178/178 tests, fmt clean, clippy identical to main, full container suite green with a genuinely-enabled `no_fallback` dead-bus run.

